PR #10, opened 2020-07-13
Remove "Capital One" logo

kalexey89:


Valve:
@kalexey89 please use netlify deploy preview to see your changes live. You can see that the logos are skewed now.
You also need to update the grid classes to make them look good

PR #18, opened 2020-10-16
Update the links and code snippets for FingerprintJS v3

Finesse:
The demo page script source code is taken from https://github.com/spalt08/fingerprintjs.com. The changes of the demo source code are located at https://github.com/spalt08/fingerprintjs.com/pull/1.


legacy/paywall.html
```diff
@@ -5,7 +5,6 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <meta name="description" content="Browser Fingerprinting, JavaScript identification, fraud detection">
-  <link rel="preconnect" href="https://cdn.fpjs.io">
```
Finesse, line 8: It makes no sense because the CDN is connected immediately (a few lines below)

spalt08 (approved):
manually tested with deploy preview
https://5f8cf007baf0e8000794d604--fingerprintjscom.netlify.app/

legacy/paywall.html
Valve: We better keep it in the docs as is, because for heavy pages the time between preconnect and actual connect is significant and browsers can fully open the connection in the preconnect step.

PR #23, opened 2020-11-04
feat: add a link to "careers" in the footer

tan-90:
Added a "Careers" link to the footer that opens `https://fingerprintjs.breezy.hr/` in a new tab.

### DIFF
--- a/src/data/site.json
+++ b/src/data/site.json
@@ -39,6 +39,10 @@
           "url": "https://dev.fingerprintjs.com/docs/privacy-policy"
         },
         {
+          "title": "Careers",
+          "url": "https://fingerprintjs.breezy.hr/"
+        },
+        {
           "title": "Log In",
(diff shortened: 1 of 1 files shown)


src/data/site.json
```diff
@@ -38,6 +38,10 @@
           "title": "Privacy Policy",
           "url": "https://dev.fingerprintjs.com/docs/privacy-policy"
         },
+        {
+          "title": "Careers",
+          "url": "https://fingerprintjs.breezy.hr/"
```
Valve, line 43: this should be http, we don't have this page on https yet.

Finesse: It works with [https](https://fingerprintjs.breezy.hr/) on my machine

Valve: oh, sorry, I meant that our cname'd version is HTTP only: http://careers.fingerprintjs.com/
Let's use the non-cnamed URL.

PR #38, opened 2020-11-20
feat: migration to Gatsby + Netlify CMS (#38)

Lisenish:
Migration to the Gtasby + new demo page + CMS integration

Co-authored-by: tan-90 <tan-90@outlook.com>
Co-authored-by: Boris Lobanov <boris.lobanov@fingerprintjs.com>


spalt08:
Questions and suggestions:
- List the environmental variables for project
- Why do we need `static/img` and `src/img` folders? Can we merge them and use only one? I suggest to have only src folder without static, since all assets are static and can be preprocessed in future (png compression for example)
- Do we need contexts if we are using hooks? Both forms and FPJS can be wrapped with hooks
- Can we move templates under cms folder to have a clear structure?
- What the reason to have equal files? (browser.ts, detector.ts under utils)
- Why vendor css files are not exported from the nam packages?
- I suggest to name component files as `Component.tsx` not `index.tsx`. This will be more convenient since IDE tab names display only filename, and the filename is clear inside IDE file tree
- Components folder structure is unclear for me. Common folder looks more like UI folder to me, and other components looks like common ones

Lisenish:
@spalt08 

- Agree, I also wanted to brush up README, will do
- It's a Gatsby thing, the ones in the static folder is raw images without any preprocessing. And yes it's a mistake that many of them in the static folder we'll fix it. And I also want to organize images better in this folder, but we still need both folders for edge-cases.
- There is no contradiction between hooks and context they can be used together. I suppose you mean that they could use a closure for caching but if two components use the same hook and only one of them update the value the other one should re-render too, at it wouldn't work with a simple closure. Probably for the form component, it's overcomplication, we'll discuss this. I also agree that they should be wrapped in hooks and provide API in form of the hooks (FPJS already does that, we will fix Forms context).
- Yeah I think we can, agree, we will move it.
- Equal files were there because they were implemented in different branches and merged simultaneously. I've fixed that thanks for noticing 
- Swiper styles are exported about others - good point, I missed that, will discuss with the team.
- It's a controversial topic, let's discuss in js channel our common approach I'm fine with both. I'll post a discussion thread today
- Yep, folders structure is a mess now, it's WIP. I have the plan to organize it better this week

PR #40, opened 2020-11-23
breakpoint variables + rollbar plugin

borislobanov92:


content/pages/account-sharing-test-page.md
```diff
@@ -1,11 +1,11 @@
 ---
 templateKey: account-sharing-page
-title: Cats are great!
+title: Dogs are great!
```
Lisenish, line 3: Oh, it shows this change because develop if off the master and you probably branched from the master? 
You should have used a develop branch as parent

src/styles/vendors/code-theme.scss
```diff
@@ -212,3 +212,7 @@ pre[class*='language-'].line-numbers > code {
   padding-right: 2.5rem !important;
   text-align: right;
 }
+
+.toolbar {
+  display: none;;
```
Lisenish, line 217: Minor: two semis and no new line at EOF

src/styles/_variables.scss
```diff
@@ -13,4 +13,6 @@ $brand-colors: (
   'brand-water': hsl(234, 52%, 11%),
 );
 
-$grid-base: 8px;
\ No newline at end of file
+$grid-base: 8px;
+
+$breakpoints: (phone: 420px, phoneLandscape: 640px, tablet: 768px, tabletWide: 860px, desktop: 1024px);
```
Lisenish, line 18: They have media-expressions for such things as landscape, wouldn't it work here? (https://eduardoboucas.github.io/include-media/documentation/#variable-media-expressions)

src/hooks/useFetch.ts
```diff
@@ -63,6 +63,6 @@ export function useFetch<T = unknown>(url?: string, options?: AxiosRequestConfig
     return () => {
       cancelRequest = true
     }
-  }, [url]) //TODO: [DI]: Fix deps
```
Lisenish, line 66: Not sure if this is a right fix, probably need to memo with deep comparison for options since clients of this hook likely to create this config object inline in a render function.

Check if we now do multiple queries because of that fix now

Lisenish: I just saw that you decided to handle it in client code. That works then, but it's error-prone, we need to constantly remember about this when using this hook, not very convenient and looks like leaking abstraction. I would prefer to handle that internally

src/helpers/rollbar.ts
```diff
@@ -0,0 +1,18 @@
+import { useState } from 'react'
+import { GATSBY_ROLLBAR_ACCESS_TOKEN } from './../constants/env'
+import Rollbar from 'rollbar'
+
+export default function useRollbar(): Rollbar {
+  const [rollbar] = useState(
```
Lisenish, line 6: Is it fine that we re-init it in each component that uses it?
Can you test that it doesn't cause some undesirable side-effects? (e.g. double posting error if there are two components on a page which use this hook)

src/components/FpjsWidget/index.tsx
```diff
@@ -40,9 +43,7 @@ export default function FpjsWidget() {
           setCurrentVisit(visits[0])
         }
       } catch (e) {
-        //TODO: [DI]: Add Rollbar report
-        // eslint-disable-next-line no-console
-        console.error(`Fingerprint loading failed: ${e.message}`)
+        rollbar.error('Unable to initialize FingerprintJS Pro', e)
```
Lisenish, line 46: Thank you 👍

src/components/FpjsWidget/FpjsWidget.module.scss
```diff
@@ -425,16 +434,20 @@
...
     display: none;
   }
 }
 
 .mobileOnly {
-  @media screen and (min-width: 641px) {
+  @include media(">phoneLandscape") {
     display: none;
   }
 }
```
Lisenish, line 453: Please install editorconfig extension to automatically insert line breaks at eof.

src/components/FpjsWidget/FpjsWidget.module.scss
```diff
@@ -425,16 +434,20 @@
...
   top: 0;
+
+  @include media(">phoneLandscape") {
+    height: 75rem;
+  }
 }
 
 .desktopOnly {
-  @media screen and (max-width: 641px) {
+  @include media("<=phoneLandscape") {
```
Lisenish, line 444: Looks like it should be vice versa?

Lisenish (changes requested):
Nothing major, please take a look at comments

src/components/FpjsWidget/FpjsWidget.module.scss
borislobanov92: it is confusing but it's actually correct: `desktopOnly` is a marker class for things that should have `display:none` on mobile screen size and vice versa

src/styles/_variables.scss
borislobanov92: not applicable here

src/hooks/useFetch.ts
borislobanov92: decided to leave it as is for now

src/components/FpjsWidget/FpjsWidget.module.scss
borislobanov92: done

content/pages/account-sharing-test-page.md
borislobanov92: accidentally pulled master instead of develop branch, won't do it in the future

Lisenish (approved):
LGTM
There're still some issues on preview, let's fix them and merge (I DMed you about them)

PR #42, opened 2020-11-25
Feature/widget card grid

tan-90:


content/pages/account-sharing-test-page.md
```diff
@@ -6,6 +6,7 @@ block1:
   bullets:
     - paws
     - tail
-    - boop
```
tan-90, line 9: This was removed on accident

Lisenish (changes requested):

Lisenish:
@tan-90 Could you please recreate whole component on the test page? We should have not only Cards Grid but this title and subtitle also, we discussed that we need to create this more complex component from simpler ones. And on the test page use the same copy and icons to see how it'll look like with real usage data.
image

src/components/widgets/CardGrid/CardGrid.module.scss
```diff
@@ -0,0 +1,46 @@
+.card {
+  flex: 1 0 auto;
```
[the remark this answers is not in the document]
tan-90: This is to ensure the height stays the same, so I think so. Open to discuss if that's wrong though.

src/templates/account-sharing-page.tsx
```diff
@@ -33,19 +39,32 @@ export const pageQuery = graphql`
             }
           }
         }
+        cards {
+          icon {
+            childImageSharp {
+              fluid(maxWidth: 2048, quality: 100) {
```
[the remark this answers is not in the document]
tan-90: That was my bad. I copied the image query from another page and forgot to adjust.

src/templates/account-sharing-page.tsx
```diff
@@ -68,3 +88,14 @@ function mapToBlockWithImage(queryBlocK: QueryBlock): BlockWithImage {
     subTitle: queryBlocK?.subheader ?? 'Default',
   } as BlockWithImage
 }
+
+type QueryCard = NonNullable<NonNullable<GatsbyTypes.AccountSharingPageQuery['markdownRemark']>['frontmatter']>['cards']
```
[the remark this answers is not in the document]
tan-90: From `gatsby-types.d.ts`: `readonly cards: Maybe<ReadonlyArray<Maybe<MarkdownRemarkFrontmatterCards>>>;`.

It returns an array, but the actual query type is an array already, correct?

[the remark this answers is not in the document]
Lisenish: 1. I meant currently this type called like it should represent a singular object but it actually represents an array.
Since it's stub page anyway we can leave it as is, but currently, it also shows types error here, does `yarn lint` work for you? Pls fix it, since it's temporary code anyway you can just type cast it if there is a need
And no `frontmatter` type is not an array, it's metadata of md file, one md file has only one frontmatter meta-data.

2. Can you add https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md this rule as a warning to eslint and fix it below? You will need also add this override to eslintrc.js
 ```
   {
      files: ['*.ts', '*.tsx'],
      parserOptions: {
        project: './tsconfig.json',
      },
    },
```

File: src/components/widgets/CardSection/CardSection.module.scss [hunk 5]
```diff
@@ -0,0 +1,21 @@
+.cardSection {
+  background-color: rgb(242, 243, 247);
+  padding: 8rem;
+  padding-top: 4rem;
+  padding-bottom: 0;
+
+  margin-top: 2rem;
+  margin-bottom: 2rem;
+
+  background-image: url('./CardSectionBackground.png');
```
Lisenish, line 10: Let's use a common naming convention for assets that are not styles, it's usually kebab-case (as we have in other places)

src/components/widgets/SubHeader/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+export interface SubHeader {
+  title: string
+  subtitle?: string
+}
+
+export default function SubHeaderComponent({ title, subtitle }: SubHeader) {
+  return (
+    <header className={styles.subHeader}>
+      <h2 className={styles.title}>{title}</h2>
+      {subtitle && <h3 className={styles.subtitle}>{subtitle}</h3>}
```
Lisenish, line 14: I thought it should be different components for easier composition, or maybe even one component but with different variants for `variant`/`size` prorerty (e.g. https://material-ui.com/ru/components/typography/, https://chakra-ui.com/docs/typography/heading)

src/components/widgets/SubHeader/SubHeader.module.scss
```diff
@@ -0,0 +1,16 @@
+.subHeader {
+  .title {
+    text-align: center;
```
Lisenish, line 3: I think alignment probably should be either customizable or set up on container level styles or this will be less reusable

src/components/widgets/SubHeader/index.tsx
Lisenish: Also, you need to replace similar text in other components to use this component ( I think we discussed in Slack that Hero component uses something similar). 
So it will be easier for you to understand how you should implement it so it's reusable when you already have different usage places for it

Lisenish (changes requested):

Lisenish:
@tan-90 Final fixes and let's merge. Or for simplicity, we can even merge it as is and you can fix final comment in different WIP branch

src/components/widgets/SubHeader/index.tsx
tan-90: Didn't change the header on Hero since the font sizes were different. I'll look through everything on Figma and on our components to see what should be customizable.

src/templates/account-sharing-page.tsx
```diff
@@ -16,9 +17,10 @@ export default function AccountSharingPage({ data }: { data: GatsbyTypes.Account
     !data.markdownRemark?.frontmatter?.block2 ||
     !data.markdownRemark?.frontmatter?.subHeader ||
     !data.markdownRemark?.frontmatter?.cards ||
-    !data.markdownRemark?.frontmatter?.cardSection
+    !data.markdownRemark?.frontmatter?.cardSection ||
+    !data.markdownRemark?.frontmatter?.inlineCta
   ) {
-    return
+    return <></>
```
Lisenish, line 23: You can return null for such cases

tan-90: I remember it didn't build at some point because of this, so I changed it. Will check and remove if unecessary.

tan-90: Breaks my build: https://reactjs.org/docs/error-decoder.html/?invariant=152&args[]=CmsStaticPage

PR #43, opened 2020-11-25
Added contact sales form.

tan-90:


Lisenish (changes requested):
LGTM but we need to wait when API is fixed before the merge

src/hooks/useForm.ts
```diff
@@ -0,0 +1,22 @@
+import { useContext } from 'react'
+import FormContext, { Forms, UPDATE_ERROR_MESSAGE, UPDATE_FORM_STATE } from '../context/FormContext'
```
Lisenish, line 2: I think we don't need these separate context file, it's all could be here. But let's do it later

src/context/FormContext.tsx
```diff
@@ -1,43 +1,72 @@
-import React, { createContext, useState } from 'react'
+import React, { createContext, Dispatch, Reducer, ReducerAction, useReducer } from 'react'
 import { FormState } from '../types/FormState'
 
+export const UPDATE_FORM_STATE = 'UPDATE_FORM_STATE'
+export const UPDATE_ERROR_MESSAGE = 'UPDATE_ERROR_MESSAGE'
+
+export enum Forms {
+  Signup = 'Signup',
```
Lisenish, line 8: I think it would be ok to have regular enum without string values

src/context/FormContext.tsx
```diff
@@ -1,43 +1,72 @@
-import React, { createContext, useState } from 'react'
+import React, { createContext, Dispatch, Reducer, ReducerAction, useReducer } from 'react'
 import { FormState } from '../types/FormState'
 
+export const UPDATE_FORM_STATE = 'UPDATE_FORM_STATE'
```
Lisenish, line 4: Maybe that's also should be an enum?

src/context/FormContext.tsx
```diff
@@ -1,43 +1,72 @@
...
   }
+}
 
-  const updateErrorMessage = (message: string) => {
-    setErrorMessage(message)
-  }
+export function FormProvider({ children }: { children: React.ReactNode }) {
+  const initialState = {}
 
+  Object.keys(Forms).forEach((form) => {
```
Lisenish, line 51: Probably it's better to treat an empty value as a Default state

src/hooks/useForm.ts
```diff
@@ -0,0 +1,22 @@
...
+import { FormState } from '../types/FormState'
+
+export default function useForm(formId: Forms) {
+  const { state, dispatch } = useContext(FormContext)
+
+  const updateFormState = (formState: FormState) => {
```
Lisenish, line 8: I think in custom hook, it's better to return memoized functions (use `useCallback`)

src/context/FormContext.tsx
tan-90: Changing this to an enum makes TS complain about the type of payload for some reason. I changed it, but we should probably discuss this on a call since it requires some extra changes.

src/context/FormContext.tsx
tan-90: You mean removing the `Default` entry on the enum and replace it with an empty value?

PR #45, opened 2020-11-27
feat: alternating image component

borislobanov92:


src/components/common/TextList/TestList.module.scss
```diff
@@ -7,9 +7,12 @@
     &:before {
```
Lisenish, line 7: Minor: Probably it's my mistake, but I've noticed that this file is called `TestList`

src/components/widgets/AlternatingImagesText/AlternatingImagesText.module.scss
```diff
@@ -0,0 +1,50 @@
+.root {
```
Lisenish, line 1: Thank you for rename :)

static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: feature/altrenating-image-component
```
Lisenish, line 3: Don't forget to rollback before the merge

src/components/widgets/AlternatingImagesText/index.tsx
```diff
@@ -16,26 +24,34 @@ export interface AlternatingImagesTextProps {
...
+    <Container className={styles.container}>
       <PreviewCompatibleImage className={styles.image} imageInfo={block.image} />
 
-      <div className={styles.text}>
+      <div className={classNames(styles.text, { [styles.left]: block.isImageAfterText })}>
         <h3 className={styles.blockTitle}>{block.subTitle}</h3>
-        <TextList items={block.bullets} ordered />
+        <TextList className={styles.list} items={block.bullets} ordered />
+        {block.ctaText &&
+          (block.isCtaButton ? (
```
Lisenish, line 46: Probably it would be more readable if move the ternary part  to a const

Lisenish (approved):
Nothing major, so approve in advance. 
After fixes, you can merge into develop

src/components/widgets/AlternatingImagesText/index.tsx
```diff
@@ -16,27 +24,42 @@ export interface AlternatingImagesTextProps {
...
 }
 
-//TODO: [DI]: Final fixes for styles, add ability to switch picture side in block
+interface CtaComponentProps {
+  ctaText: string
+  ctaUrl: string
+  isCtaButton: boolean
+}
+
+function CtaComponent({ ctaUrl, ctaText, isCtaButton }: CtaComponentProps) {
```
Lisenish, line 57: I meant that you can just extract this to something like `const ctaElement = isCtaButton ?  ... : ...` not exactly as a separate component. But this works too.

PR #46, opened 2020-11-27
Feature: inline CTA widget

tan-90:
Also fixed a padding issue on the signup form, removed the `GATSBY_` prefix from the environment constants, and addressed the form context PR change requests.


src/components/ContactSalesForm/index.tsx
```diff
@@ -5,7 +5,7 @@ import { ReactComponent as CloseSvg } from '../../img/close.svg'
 import { FormState } from '../../types/FormState'
 import { sendEvent } from '../../helpers/gtm'
 import classNames from 'classnames'
-import { Forms } from '../../context/FormContext'
+import { Forms } from '../../hooks/useForm'
```
Lisenish, line 8: Why we have two separate imports from /useForm?

src/components/FpjsWidget/MobileWidget.tsx
```diff
@@ -10,9 +10,9 @@ import { getBotDecision, getBrowserName, getVisitTitle } from '../../helpers/fpj
 import { ReactComponent as InfoSvg } from './info.svg'
 import Tippy from '@tippyjs/react'
 import styles from './MobileWidget.module.scss'
-import { GATSBY_MAPBOX_ACCESS_TOKEN } from '../../constants/env'
+import { MAPBOX_ACCESS_TOKEN } from '../../constants/env'
 
-const mapboxToken = GATSBY_MAPBOX_ACCESS_TOKEN
+const mapboxToken = MAPBOX_ACCESS_TOKEN
```
Lisenish, line 15: We can inline it

src/components/FpjsWidget/index.tsx
```diff
@@ -9,12 +9,12 @@ import { CurrentVisitProps } from './currentVisitProps'
...
 import { useVisitorData } from '../../context/FpjsContext'
 import useRollbar from '../../hooks/useRollbar'
-import { GATSBY_FPJS_API_TOKEN, GATSBY_FPJS_ENDPOINT, GATSBY_MAPBOX_ACCESS_TOKEN } from '../../constants/env'
+import { FPJS_API_TOKEN, FPJS_ENDPOINT, MAPBOX_ACCESS_TOKEN } from '../../constants/env'
 import styles from './FpjsWidget.module.scss'
 
-const apiToken = GATSBY_FPJS_API_TOKEN ?? 'test_fpjs_api_token'
-const endpoint = GATSBY_FPJS_ENDPOINT ?? ''
-const mapboxToken = GATSBY_MAPBOX_ACCESS_TOKEN
+const apiToken = FPJS_API_TOKEN ?? 'test_fpjs_api_token'
```
Lisenish, line 15: Can you put these ?? checks for default values inside ../../constants/env? So we don't need to put it in every usage place

src/components/GetStartedForm/index.tsx
```diff
@@ -5,10 +5,10 @@ import { ReactComponent as CloseSvg } from '../../img/close.svg'
 import classNames from 'classnames'
 import Button from '../common/Button'
 import { FormState } from '../../types/FormState'
-import { GATSBY_FPJS_DASHBOARD_ENDPOINT } from '../../constants/env'
+import { FPJS_DASHBOARD_ENDPOINT } from '../../constants/env'
 import { useVisitorData } from '../../context/FpjsContext'
 import { sendEvent } from '../../helpers/gtm'
-import { Forms } from '../../context/FormContext'
+import { Forms } from '../../hooks/useForm'
```
Lisenish, line 11: Multiple imports of the same file

src/components/demo/SignupSection/SignupSection.module.scss
```diff
@@ -1,6 +1,6 @@
 .section {
   padding-top: 2rem;
-  padding-bottom: 2rem;
+  padding-bottom: 0;
```
Lisenish, line 3: - I'm suspicious why this padding was there in the first place... Have you checked all usage places? Maybe this change fix one place and breaks the other
- You probably can just delete it and not set to 0?

src/components/widgets/InlineCta/InlineCta.module.scss
```diff
@@ -0,0 +1,70 @@
+.inlineCta {
```
Lisenish, line 1: A common practice to call this class .root for the component's root element but up to you

src/components/widgets/InlineCta/InlineCta.module.scss
```diff
@@ -0,0 +1,70 @@
...
+  position: relative;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  overflow-y: hidden;
+
+  margin: 2rem 0 2rem 0;
```
Lisenish, line 8: You can use short notation with two values for that (same for padding)

src/components/widgets/InlineCta/InlineCta.module.scss
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @include media('<=phoneLandscape') {
+    flex-direction: column;
+    justify-content: flex-start;
+
+    padding: 4rem 3rem 4rem 3rem;
+  }
+
+  .header {
```
Lisenish, line 38: Looks familiar to me, need to merge in develop both your branches and use our header component here

src/hooks/useForm.tsx
```diff
@@ -1,12 +1,31 @@
-import React, { createContext, Dispatch, Reducer, ReducerAction, useReducer } from 'react'
+import React, { createContext, Dispatch, Reducer, ReducerAction, useReducer, useContext, useCallback } from 'react'
 import { FormState } from '../types/FormState'
 
+export default function useForm(formId: Forms) {
```
Lisenish, line 4: It shouldn't be a default export, let's leave it as regular named export

Lisenish (changes requested):
@tan-90 Please sync with develop branch. Don't spend too much time on merging temp data (account-sharing-page and it's settings)

src/components/widgets/InlineCta/InlineCta.module.scss
tan-90: I prefer using the name, but I think it's best to stick to the convention, I'll change it.

src/components/demo/SignupSection/SignupSection.module.scss
tan-90: I am using the section component, which has some padding by default. It allows passing in a class, which is what I did since this section didn't need padding. Here's how it looks without it:

image

PR #48, opened 2020-12-03
Feature/tech fixes

Lisenish:


src/hooks/useBackgroundImage.ts
```diff
@@ -0,0 +1,21 @@
+import { graphql, useStaticQuery } from 'gatsby'
+
+export function useBackgroundImage() {
```
borislobanov92, line 3: should this hooks perhaps have a parameter with the image name of file path?

borislobanov92 (approved):

tan-90 (approved):

src/hooks/useBackgroundImage.ts
Lisenish: Yeah, you right, forgot to implement it.

Actually, the plan was to call it `useMainBackroundImage` and extract private reusable hook, so we don't need to copy filename in each usage place. 

I will do it

Lisenish: I've remembered why I haven't implemented that, string interpolation doesn't work with graphql tagged templates.
Also, I've thought to implement that with fragments, but parameterized fragments are not supported by GraphQL

So I believe the only option now to copy-paste similar code for different bg's.
I've renamed hook so it doesn't look like it's a generic one

PR #49, opened 2020-12-07
Added blog pages with long-form content template

tan-90:


content/blog/blog-test-page.md
```diff
@@ -0,0 +1,58 @@
+---
```
Lisenish, line 1: Let's delete this one before the merge.

I will set up storybook this or next week so we don't have to create temp pages to check markup (but locally probably we still have to create the to check generation)

src/pages/blog/blog.module.scss
```diff
@@ -0,0 +1,83 @@
+.header {
```
Lisenish, line 1: Shouldn't we use our typography header component instead?

src/pages/blog/blog.module.scss
```diff
@@ -0,0 +1,83 @@
...
+  box-shadow: 0px 0px 20px 20px rgba(72, 76, 108, 0.044116);
+  background-color: v('white');
+
+  transition-duration: 0.5s;
+
+  .title {
+    transition-duration: 0.5s;
+  }
+
+  &:hover {
```
Lisenish, line 50: I love that you added this nice little touch :) (it's probably from a reference template?)

src/pages/blog/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    <Layout>
+      <Section>
+        <Container size='large'>
+          <h1 className={styles.header}>Blog landing page</h1>
+
+          <Section className={styles.grid}>
+            {posts.map(({ node: post }) => (
+              <div key={post.id} className={styles.item}>
+                <Post
+                  title={post.frontmatter?.metadata?.title ?? ''}
```
Lisenish, line 23: Looks a little dirty, you can destruct all fields to variables (need to convert arrow function to multi-line)

gatsby-config.js
```diff
@@ -40,6 +40,13 @@ module.exports = {
         name: 'pages',
       },
     },
+    {
+      resolve: 'gatsby-source-filesystem',
```
Lisenish, line 44: Do we really need to register additional folder if we already registered content?
I think it should pick up sub-folders

static/admin/config.yml
```diff
@@ -105,3 +105,23 @@ collections:
                   { label: Button href, name: buttonHref, widget: string },
                 ],
             }
+  - name: 'blog'
+    label: 'Blog'
+    folder: content/blog
+    create: true
```
Lisenish, line 111: Probably should allow deleting also? Not 100% sure, we may ask Savannah

src/templates/long-form-content.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { baseUrl } from '../constants/content'
+
+import styles from './long-form-content.module.scss'
+
+const HtmlContent = ({ content, className }: { content: string; className?: string }) => (
+  <div className={className} dangerouslySetInnerHTML={{ __html: content }} />
+)
+
+const Content = ({ content, className }: { content: string | JSX.Element; className?: string }) => (
+  <div className={className}>{content}</div>
```
Lisenish, line 16: Maybe use `children` here instead of a specific named prop?

src/templates/long-form-content.tsx
```diff
@@ -0,0 +1,90 @@
...
+        }
+      }
+    }
+  }
+`
+
+interface TemplateProps {
+  metadata: GatsbyTypes.SiteSiteMetadata
+  body: string | JSX.Element
+  contentComponent?: (props) => JSX.Element
```
Lisenish, line 56: Same here: maybe use `children` here instead of a named prop? Also, it's incorrect typing for a component probably you need just `React.ReactNode` (same for body above)

Lisenish (changes requested):

Lisenish:
@tan-90 Pls sync this branch with develop

src/pages/blog/blog.module.scss
borislobanov92: ++

src/pages/blog/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+export default function Blog({ data }: { data: GatsbyTypes.BlogQuery }) {
+  const { edges: posts } = data.allMarkdownRemark
+
+  return (
+    <Layout>
+      <Section>
+        <Container size='large'>
+          <h1 className={styles.header}>Blog landing page</h1>
+
+          <Section className={styles.grid}>
```
borislobanov92, line 19: I think sections should only be on 1 level, breaking the page content into horizontal blocks. I suggest that you move the h1 to the top of Layout content and keep just the upper level section

src/pages/blog/blog.module.scss
```diff
@@ -0,0 +1,83 @@
...
+  font-weight: v('bold');
+  font-size: 4rem;
+  padding-bottom: 2rem;
+}
+
+.grid {
+  padding: 3rem 0;
+
+  display: flex;
+  flex-wrap: wrap;
```
borislobanov92, line 12: tbh, it's kinda strange seeing a class named 'grid' with display 'flex'. I would consider using `display: grid` since that's basically what it was created for. this is of course an optional fix but have you tested this with multiple posts? is everything working properly on all screen resolutions?

src/templates/long-form-content.module.scss
```diff
@@ -0,0 +1,139 @@
+.root {
+  .container {
```
borislobanov92, line 2: I don't think we should nest these since it doesn't help us in any way and only generates longer class names and makes it harder to debug in the inspector

src/templates/long-form-content.module.scss
```diff
@@ -0,0 +1,139 @@
...
+
+    .title {
+      flex-basis: 40%;
+      font-size: 2.75rem;
+      font-family: v('ff-mono');
+
+      padding-right: 8rem;
+
+      @include media('<=tabletWide') {
+        margin-bottom: 6.75rem;
```
borislobanov92, line 19: looks too big on mobile

borislobanov92 (changes requested):

src/templates/long-form-content.module.scss
Lisenish: Agreed, we even discussed that in some previous PR before with @tan-90. 
Obviously, we need 1 level nesting for elements rules but all other nestings seems excessive

It also leads to some bugs with the background-image plugin we use (it's actually their bug but still)

static/admin/config.yml
tan-90: That was an oversight yeah, I feel like it would be a problem if all created pages were just permanent.

gatsby-config.js
tan-90: I got confused with the name and thought we needed one for each subfolder. It does pick everything up without this.

src/templates/long-form-content.module.scss
tan-90: Not nesting unless it's for element rules is a very good guideline, thank you. I'll start following that.

src/templates/long-form-content.tsx
tan-90: Here I don't think we can, because this needs to have the same signature as the `HtmlContent` component so we can pick `HtmlContent` for the actual page and `Content` for the preview. `HtmlContent` expects a string so it can use `dangerouslySetInnerHTML`, which wouldn't work with a ReactNode. Maybe there is a way to do it though, open to suggestions.

I did change the type from `JSX.Element` to `ReactNode`.

src/templates/long-form-content.tsx
tan-90: I changed the type of `body` to `string | React.ReactNode`, but I think it's the same as before, `body` can't be just `children: React.ReactNode` because the template doesn't know which `contentComponent` it's gonna use, and the `HtmlContent` requires a string.

As for the `contentComponent` type, I'm using it on lines 59 and 67 to create a component, not just to render a node. TS complains if the return type is not a `JSX.Element`.

I took this from the starter and added the typing. The `contentComponent` props were missing a type, so I fixed that.

src/templates/long-form-content.tsx
tan-90: Unless `children` could be `string | React.ReactNode`. But even if that's possible, I don't think it's a good idea.

src/pages/blog/index.tsx
tan-90: That was my bad yeah, the header is part of the section. Just changed it to be a single section.

src/pages/blog/blog.module.scss
tan-90: Thanks! It's from here https://www.fastly.com/blog. I wanted to get that "lifting off the background" effect.

src/pages/blog/blog.module.scss
tan-90: The mockup on Figma is just slightly smaller than our typography header. But I do agree that it's better to be consistent, I'll change that.

PR #54, opened 2020-12-15
Blog collection

tan-90:
Added long-form content component.
Added blog collection and the ability to create posts.
Added blog landing page.


Lisenish (approved):
@tan-90 LGTM! Let's ask Savannah to fill out the first pages.

We also need to discuss where to put a link to blog landing page (maybe we this is related to to nav task)

PR #55, opened 2020-12-15
Blog pagination

tan-90:
Adds blog pagination.

Discussed with @Lisenish and we decided it's ok to branch from feature branches to keep tickets separate, so this depends on https://github.com/fingerprintjs/fingerprintjs.com/pull/54 being merged.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -58,7 +58,7 @@ exports.createPages = async ({ actions, graphql }) => {
   const postsPerPage = 12
   const numPages = Math.ceil(blogPosts.length / postsPerPage)
 
-  Array.from({ length: numPages }).forEach((_, i) => {
+  for (let i = 0; i < numPages; ++i) {
     createPage({
       // The first page doesn't need a number.
       path: `blog${i === 0 ? '/' : `/${i + 1}/`}`,
@@ -70,7 +70,7 @@ exports.createPages = async ({ actions, graphql }) => {
         currentPage: i + 1,
       },
     })
-  })
+  }
 }
 
 function createNodePath({ node, getNode }) {
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -1,12 +1,13 @@
 const path = require('path')
 const { createFilePath } = require('gatsby-source-filesystem')
 
-exports.createPages = async ({ actions, graphql }) => {
-  const { createPage } = actions
-
+async function getFolderEdges(folder, graphql) {
   const { data, errors } = await graphql(`
     {
-      allMarkdownRemark(limit: 1000) {
+      allMarkdownRemark(
+        limit: 1000
+        filter: {fileAbsolutePath: {regex: "/(${folder})/.*\.md$/"}}
+      ) {
         edges {
           node {
             id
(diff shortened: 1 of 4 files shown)


gatsby-node.js
```diff
@@ -1,12 +1,13 @@
...
-exports.createPages = async ({ actions, graphql }) => {
-  const { createPage } = actions
-
+async function getFolderEdges(folder, graphql) {
   const { data, errors } = await graphql(`
     {
-      allMarkdownRemark(limit: 1000) {
+      allMarkdownRemark(
+        limit: 1000
+        filter: {fileAbsolutePath: {regex: "/(${folder})/.*\.md$/"}}
```
tan-90, line 9: Filtering comes from here: https://stackoverflow.com/a/58019738
This appears to be the only way to do it.

gatsby-node.js
```diff
@@ -28,18 +29,45 @@ exports.createPages = async ({ actions, graphql }) => {
...
+  })
+}
+
+exports.createPages = async ({ actions, graphql }) => {
+  const { createPage } = actions
+
+  const pages = await getFolderEdges('index', graphql)
+  pages.forEach((edge) => createPageFromEdge(edge, createPage))
+
+  const blogPosts = await getFolderEdges('blog', graphql)
```
tan-90, line 55: We need to get the files from each folder separately, otherwise we won't know how many blog posts there are. This does mean adding new folders will require them being added here as well, so that's something to keep in mind.

gatsby-node.js
```diff
@@ -28,18 +29,45 @@ exports.createPages = async ({ actions, graphql }) => {
...
+  blogPosts.forEach((edge) => createPageFromEdge(edge, createPage))
 
-  posts.forEach((edge) => {
-    const id = edge.node.id
+  const postsPerPage = 12
+  const numPages = Math.ceil(blogPosts.length / postsPerPage)
 
+  Array.from({ length: numPages }).forEach((_, i) => {
```
tan-90, line 61: This comes from here: https://nickymeuleman.netlify.app/blog/gatsby-pagination#create-listing-pages
I'd just go with a for loop but decided to follow the example, maybe this is better.

gatsby-node.js
```diff
@@ -28,18 +29,45 @@ exports.createPages = async ({ actions, graphql }) => {
...
     createPage({
-      path: edge.node.fields.slug,
-      tags: edge.node.frontmatter.tags,
-      component: path.resolve(`src/templates/${String(edge.node.frontmatter.templateKey)}.tsx`),
-      // additional data can be passed via context
+      // The first page doesn't need a number.
```
tan-90, line 63: This was the cleaner way I found to do this. Still didn't look clear enough so I added a comment.

src/templates/blog.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  description={description}
+                  publishDate={dateFormatter.format(new Date(publishDate))}
+                  image={image as GatsbyTypes.File}
+                  path={slug}
+                />
+              )
+            })}
+          </div>
+
+          {numPages > 1 && (
```
tan-90, line 49: If we remove this, the links wouldn't show regardless, but we need to remove the container as well, so this is necessary.

gatsby-node.js
Lisenish: I thought you said you're going to use for loop in the slack discussion :)

src/templates/blog.tsx
```diff
@@ -1,15 +1,22 @@
...
+import Layout from '../components/Layout'
 import Img from 'gatsby-image'
-import Container from '../../components/common/Container'
+import Container from '../components/common/Container'
+import classNames from 'classnames'
 
 import styles from './blog.module.scss'
 
-export default function Blog({ data }: { data: GatsbyTypes.BlogQuery }) {
+export default function Blog({ data, pageContext }: { data: GatsbyTypes.BlogQuery; pageContext: BlogContext }) {
```
Lisenish, line 11: Minor: I would extract type to the interface here for readability, it's too big to inline it with second field.

Lisenish (approved):

Lisenish:
@tan-90 Have you tested this locally? Since we gonna wait Savnnah to create the first page for a while lets test it and merge into base feature, so we don't have too many hanging branches

gatsby-node.js
tan-90: Yep, will change :)

tan-90:
> @tan-90 Have you tested this locally? Since we gonna wait Savnnah to create the first page for a while lets test it and merge into base feature, so we don't have too many hanging branches

I have tested it locally, I think it's ok to merge with `feature/blog-collection` and go on from there.

PR #57, opened 2020-12-16
Fix url slashes

natapg:
I added the URLs to the content.ts file and changed them in the index pages. We need to discuss if this way of saving URLs as constants is ok


src/components/GithubButton/index.tsx
```diff
@@ -16,7 +17,7 @@ interface GithubButtonProps {
 const githubToken = GITHUB_API_TOKEN
 
 export default function GithubButton({ className }: GithubButtonProps) {
-  const url = 'https://api.github.com/repos/fingerprintjs/fingerprintjs'
+  const url = githubApiUrl
```
Lisenish, line 20: No point in this variable now, you can inline it

src/constants/content.ts
```diff
@@ -1 +1,20 @@
 export const baseUrl = 'https://fingerprintjs.com'
+export const demoUrl = '/demo/'
```
Lisenish, line 2: Let's semantically group them in this way 
`export const URL = {
  example: 'https://example.com'
} as const`

Groups can be: DOC_URL (for documentation), PATH (for internal relative links), URL(for external urls), MAILTO (for one mailto link)
In that case we even don't need url postfixes

Also let's rename baseUrl to BASE_URL.

src/constants/content.ts
```diff
@@ -1 +1,20 @@
...
+export const pricingUrl = '/pricing/'
+export const githubRepoUrl = 'https://github.com/fingerprintjs/fingerprintjs'
+export const githubApiUrl = 'https://api.github.com/repos/fingerprintjs/fingerprintjs'
+export const documentationUrl = 'https://dev.fingerprintjs.com'
+export const mailToUrl = 'mailto:support@fingerprintjs.com'
+export const dashboardLoginUrl = 'https://dashboard.fingerprintjs.com/login'
+export const termOfUseUrl = 'https://dev.fingerprintjs.com/docs/terms-of-service'
+export const privacyPolicyUrl = 'https://dev.fingerprintjs.com/docs/privacy-policy'
+export const careersUrl = 'https://fingerprintjs.breezy.hr/'
+export const getStartedUrl = 'https://dev.fingerprintjs.com/docs/introduction'
```
Lisenish, line 12: We discussed that we can change it in this branch

Lisenish (changes requested):

src/components/GithubButton/index.tsx
```diff
@@ -28,7 +29,7 @@ export default function GithubButton({ className }: GithubButtonProps) {
   const { data } = useFetch<GithubReposResponse>(url, options)
```
tan-90, line 29: Could inline `githubApiUrl` here.

Lisenish:
@natapg I see that you skipped e-commerce card link. This one is trickier because it's generated, try to figure out how to fix it, but if you're stuck ask @tan-90 for help

src/constants/content.ts
```diff
@@ -1 +1,20 @@
 export const baseUrl = 'https://fingerprintjs.com'
```
tan-90, line 1: Shouldn't all of these have a trailing slash? @Lisenish

tan-90 (commented):
Looks good to me. Just need @Lisenish confirmation about trailing slashes.

src/constants/content.ts
Lisenish: Only internal ones, it doesn't matter for external (non-website) urls.

A little context: the reason we need trailing slashes because of this Netlify behaviour:

> Our CDN edge nodes do URL normalization before the redirect rules kick in. This happens to make sure we can guarantee the highest possible cache hit rate and the absolute best performance for your site.

And redirects are bad for SEO, so the only options for us to lead to the correct url from the start.

Lisenish (approved):

PR #58, opened 2020-12-17
Static page links

tan-90:
Adds links to static pages to the footer.
Breaks up the top navbar into left and right nav.

Depends on feature/static-pages


src/components/Header/index.tsx
```diff
@@ -75,3 +90,32 @@ export default function Header() {
     </>
   )
 }
+
+interface DropdownListProps {
```
Lisenish, line 94: Let's move it to separate file it's complex enough component and it has big chunk of the styles

src/components/MobileNavbar/index.tsx
```diff
@@ -91,3 +88,35 @@ export default function MobileNavbar() {
     </>
   )
 }
+
+interface DropdownMenuProps {
```
Lisenish, line 92: Same here

src/components/Header/index.tsx
```diff
@@ -10,6 +10,9 @@ import Container from '../common/Container'
 import { isBrowser } from '../../helpers/detector'
 import Modal from '../../components/common/Modal'
 import ContactSalesForm from '../../components/ContactSalesForm'
+import { useCaseLinks } from '../../constants/content'
+import classNames from 'classnames'
+import { ReactComponent as ExpandMoreSvg } from '../../img/expand-more.svg'
```
Lisenish, line 15: Do we use it somewhere else? Maybe it should be a local icon?

src/constants/content.ts
```diff
@@ -1 +1,9 @@
 export const baseUrl = 'https://fingerprintjs.com'
+
+export const useCaseLinks = [
+  { title: 'Account Takeover', url: '/account-takeover/', isLocal: true },
```
Lisenish, line 4: `isLocal` probably excessive since it's true by default?

Lisenish (changes requested):

src/components/Navbar/index.tsx
```diff
@@ -8,28 +7,14 @@ export default function Navbar() {
...
           <a className={nav.link} href='https://dev.fingerprintjs.com'>
-            Documentation
+            Docs
           </a>
           <a className={nav.link} href='mailto:support@fingerprintjs.com'>
             Support
           </a>
           <a className={nav.link} href='https://dashboard.fingerprintjs.com/login' target='_blank' rel='noreferrer'>
-            Log In
+            Login
```
Lisenish, line 17: Is it intentional change? Maybe we should change it everywhere then? It's Log In in other places

src/constants/content.ts
tan-90: I added this was to make it compatible with the type the footer is expecting:
```
interface FooterLinkSection {
  title: string
  links: { title: string; url: string; isLocal?: boolean }[]
}
```

The footer considers `false` as the default, so changing that would risk breaking something.

tan-90: Or I could just remove it and then map each link to the expected type on the Footer. I'm ok with either solution.

src/components/Header/index.tsx
tan-90: We use it on the Collapsible as well, but it's local for the collapsible. I'll fix that.

src/components/Navbar/index.tsx
tan-90: It was written Login on the list Savannah provided, but I forgot to check other places. I'll confirm with her just to make sure.

tan-90: Discussed with Savannah, changed all occurrences to "login".

Lisenish (approved):

PR #61, opened 2020-12-21
Added breadcrumbs nav and SEO.

tan-90:
Closes WEB-61


yarn.lock
```diff
@@ -65,6 +65,27 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
+"@babel/core@^7.1.0", "@babel/core@^7.7.5":
```
Lisenish, line 68: There are a lot of changes here, have you upgraded all deps?

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+import { Link } from 'gatsby'
+import { Helmet } from 'react-helmet'
+import Container from '../common/Container'
+import { baseUrl } from '../../constants/content'
```
Lisenish, line 5: You need to sync base branch with `develop` after we merge this PR, this const is renamed there

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
```
Lisenish, line 1: We decided not to use index naming for files, let's rename it to the component name.

Also, let's separate components to two files, they're completely different component it is fine to have multiple ones in one file only if one of them is a local child component.

Lisenish (changes requested):
- You need to sync with the base branch
- I don't think we need Home root breadcrumb since the logo leads to the Home page. And even if we decide that we need it the blog landing page should have it also then.
- Not sure about the current position feels like it's too distant and unattached to the main content. Maybe we can put it above the title? (but without alignment disruption of the title and text paragraph)

yarn.lock
tan-90: Weird, just did `yarn add gatsby-plugin-breadcrumb`

PR #63, opened 2020-12-22
Generate sitemap.xlm and robots.txt

natapg:
Added sitemap.xml and robots.txt to our website using Gatsby plugins
Closes WEB-70


gatsby-config.js
```diff
@@ -13,10 +13,13 @@ module.exports = {
   siteMetadata: {
     title: 'FingerprintJS Pro - browser fingerprinting & fraud detection API',
     description: 'Stop fraud, spam, and account takeovers with 99.5% accurate browser fingerprinting as a service.',
+    siteUrl: 'https://fingerprintjs.com',
```
Lisenish, line 16: You probably can't use TS file here, I get it, but you can still move this string to the constant variable in this file

Lisenish:
Pls delete `package-lock.json` file, we use yarn

Lisenish (changes requested):

gatsby-config.js
natapg: you're right, I'm going to make the change

natapg: you're right, I'm going to move it to a constant variable

gatsby-config.js
```diff
@@ -13,10 +14,13 @@ module.exports = {
   siteMetadata: {
     title: 'FingerprintJS Pro - browser fingerprinting & fraud detection API',
     description: 'Stop fraud, spam, and account takeovers with 99.5% accurate browser fingerprinting as a service.',
-    url: 'https://fingerprintjs.com',
+    siteUrl: baseUrl,
+    url: baseUrl,
```
Lisenish, line 18: @tan-90 AFAIK It's common to have it called stiteUrl in metadata and some plugins rely on that convention, so we need to refactor url usages to use siteUrl intead (in some separate task)

Lisenish (approved):

PR #64, opened 2020-12-22
Added featured posts.

tan-90:
Added `FEATURED` switch to CMS.
Added featured posts section to blog landing page.
Added landing page for featured posts.

Closes WEB-76


Lisenish:
@tan-90 Please update sync develop branch

src/templates/blog.module.scss
```diff
@@ -1,106 +1,3 @@
-.grid {
```
Lisenish, line 1: Nice :)

src/templates/blog.tsx
```diff
@@ -2,110 +2,87 @@ import { graphql, Link } from 'gatsby'
...
-          )}
+          <PaginationNav currentPage={currentPage} numPages={numPages} basePath='/blog/' />
         </Container>
       </Section>
     </Layout>
   )
 }
 
 export const pageQuery = graphql`
+  fragment PostData on MarkdownRemarkConnection {
```
Lisenish, line 41: 👍 

I believe you can actually even separate fragment to another exported const with arbitrary name just to have only page query in pageQuery const

src/templates/blog-featured.tsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+`
+
+interface BlogFeaturedContext {
+  currentPage: number
+  numPages: number
+}
+
+const dateFormatter = new Intl.DateTimeFormat('en-US', { month: 'long', day: 'numeric', year: 'numeric' })
```
Lisenish, line 78: Just left a comment in another review that we probably should have all formatter in one common place

Lisenish (changes requested):

Lisenish:
@tan-90 I think we should delete our test post and feature some existing one, so we don't ship these test page on production :)

Lisenish (approved):

Lisenish:
@tan-90 For some reason there are old link in blog cards? Something merged incorrectly or we need another sync with develop?

tan-90:
> @tan-90 For some reason there are old link in blog cards? Something merged incorrectly or we need another sync with develop?

They're like that on develop as well. Maybe it never changed? I think me and Natanael were working in parallel at the time. I'll fix it now.

Lisenish:
@tan-90 It's good that you fixed that ones, but I actually meant blog posts cards not use cases, looks like they use old filename urls instead of metaurl

PR #67, opened 2020-12-24
Created a context for fetching github stars.

tan-90:
Closes WEB-74


src/context/GithubContext.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { createContext, useContext, useEffect, useMemo, useState } from 'react'
+import { GITHUB_API_TOKEN } from '../constants/env'
+import { URL } from '../constants/content'
+
+export interface GithubReposResponse {
+  stargazers_count: number
+}
+
+const GithubContext = createContext<{ githubData?: GithubReposResponse }>({})
```
Lisenish, line 9: Creating context is overkill for this, we should've discussed this (it's maybe even overkill for a FormContext also to be honest)

Could you please remove context but just leave custom hook `useGithubData`. 
It can cache first request in the closure outside of the function and just return it if it's set

If it's gonna be implemented like that you can also simplify the useMemo part and just inline those options const.

src/components/GithubButton/index.tsx
```diff
@@ -1,45 +1,28 @@
...
-            notation: 'standard',
-            maximumFractionDigits: 1,
-          }).format(data.stargazers_count)}
-      </div>
+      {githubData && <div className={styles.counter}>{numberFormatter.format(githubData.stargazers_count)}</div>}
     </a>
   )
 }
+
+const numberFormatter = new Intl.NumberFormat('en-US', {
```
Lisenish, line 25: Good that it's now is not recreating 👍 Maybe we even should move it to common reusable helpers?

Lisenish (changes requested):

Lisenish (approved):

PR #68, opened 2020-12-24
Post-deploy fixes for static pages and blog.

tan-90:
Changed long-form content layout to be vertical.
Increased long-form content font size.
CMS page URLs now use metadata URL instead of slug.
Replaced text with path on SVG that was being cut off.

Closes WEB-84


Lisenish (changes requested):
Could be please align title by the left side? 
Otherwise it's good

Lisenish (approved):

PR #71, opened 2020-12-28
fixing scrolling issue when menu mobile is open

natapg:
Added **position:fixed** to css to disable scrolling the document body when the mobile menu is open
Closes WEB-65

### DIFF
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -23,13 +23,9 @@ export default function Header() {
 
   useEffect(() => {
     const mobileBodyClass = 'isMobileMenuOpen'
-    if (isBrowser && isMobileMenuOpen) {
-      document.body.classList.add(mobileBodyClass)
-      document.body.style.overflow = 'hidden'
-    } else {
-      document.body.classList.remove(mobileBodyClass)
-      document.body.style.overflow = 'unset'
-    }
+    isBrowser && isMobileMenuOpen
+      ? document.body.classList.add(mobileBodyClass)
+      : document.body.classList.remove(mobileBodyClass)
   }, [isMobileMenuOpen])
 
   useEffect(() => {
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -23,9 +23,11 @@ export default function Header() {
 
   useEffect(() => {
     const mobileBodyClass = 'isMobileMenuOpen'
-    isBrowser && isMobileMenuOpen
-      ? document.body.classList.add(mobileBodyClass)
-      : document.body.classList.remove(mobileBodyClass)
+    if (isBrowser && isMobileMenuOpen) {
+      document.body.classList.add(mobileBodyClass)
+    } else {
+      document.body.classList.remove(mobileBodyClass)
+    }
   }, [isMobileMenuOpen])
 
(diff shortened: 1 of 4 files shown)


natapg:
In Android the menu is no longer scrollable with this solution, I will look for another @Lisenish

natapg:
I found out that for some reason the style wasn't applied when opening the menu so I had to add "!Important" to it to override the default style

image

image

src/styles/global-styles.scss
```diff
@@ -35,7 +35,7 @@ body {
   min-height: 100vh;
 
   &.isMobileMenuOpen {
-    overflow: hidden;
+    overflow: hidden !important;
```
Lisenish, line 38: Adding important is almost always is a bad idea and dirty fix, please figure out how to fix it properly, you probably need increase specificity of the selector (https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity)

Lisenish (changes requested):

src/styles/global-styles.scss
natapg: Thanks. I was using !Important to override the inline style

src/styles/global-styles.scss
```diff
@@ -35,7 +35,7 @@ body {
   min-height: 100vh;
 
   &.isMobileMenuOpen {
-    overflow: hidden;
+    overflow: hidden ;
```
Lisenish, line 38: Trailing whitespace

src/components/Header/index.tsx
```diff
@@ -23,9 +23,13 @@ export default function Header() {
 
   useEffect(() => {
     const mobileBodyClass = 'isMobileMenuOpen'
-    isBrowser && isMobileMenuOpen
-      ? document.body.classList.add(mobileBodyClass)
-      : document.body.classList.remove(mobileBodyClass)
+    if (isBrowser && isMobileMenuOpen) {
```
Lisenish, line 26: Discussed in DM to change Menu instead so it resotres original `body.overflow` style.

P.s. @tan-90 remember we discussed such potential bug in review? :)

Lisenish (changes requested):

src/components/Header/index.tsx
```diff
@@ -23,13 +23,9 @@ export default function Header() {
...
   useEffect(() => {
     const mobileBodyClass = 'isMobileMenuOpen'
-    if (isBrowser && isMobileMenuOpen) {
-      document.body.classList.add(mobileBodyClass)
-      document.body.style.overflow = 'hidden'
-    } else {
-      document.body.classList.remove(mobileBodyClass)
-      document.body.style.overflow = 'unset'
-    }
+    isBrowser && isMobileMenuOpen
```
Lisenish: Minor code style issue: it's better to use ternary operator only for the places where you need expression, e.g. assigning new variable `const x = condition ? 1 : 0`, because the result of this expression is actully what you need.
For the places like here, where you actually use statements (imperative modifying code, and you don't doesn't return any value) it's better to stick to regular if/else (even if they consist from one line)

Lisenish (changes requested):

Lisenish (approved):

PR #73, opened 2020-12-29
Added trigger event for google optimize

Lisenish:
- Added trigger event for google optimize


src/hooks/useSiteMetadata.tsx
```diff
@@ -7,6 +7,8 @@ const defaultMeta = {
   image: 'default',
 } as const
 
+//TODO: [DI]: Use 'siteUrl' instead of 'url' here
```
Lisenish, line 10: @tan-90 Could you please fix that one, there is a  broken type in Layout now because of it

tan-90 (approved):

PR #76, opened 2020-12-31
added learn more link on hover front page card

natapg:
Closes WEB-63

### DIFF
--- a/src/components/UseCasesSection/index.tsx
+++ b/src/components/UseCasesSection/index.tsx
@@ -92,6 +92,7 @@ function UseCase({ link, local = true, icon: Icon, title, children, large, onCli
       <div className={styles.info}>
         <h3 className={styles.title}>{title}</h3>
         <p className={styles.description}>{children}</p>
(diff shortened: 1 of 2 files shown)


src/components/UseCasesSection/index.tsx
```diff
@@ -92,6 +92,7 @@ function UseCase({ link, local = true, icon: Icon, title, children, large, onCli
       <div className={styles.info}>
         <h3 className={styles.title}>{title}</h3>
         <p className={styles.description}>{children}</p>
+        <footer className={styles.learnMore}>Learn More &gt;</footer>
```
tan-90, line 95: I like that you used a specific tag!

tan-90 (approved):
Looks good to me.

src/components/UseCasesSection/index.tsx
Lisenish: Actually not sure about that. I mean I'm all for semantics but ideally we should probably use section here then since footer should be once per section, but maybe it

PR #81, opened 2021-01-04
Static page changes

tan-90:
Closes WEB-67


src/components/widgets/CardSection/index.tsx
```diff
@@ -16,10 +16,10 @@ export interface CardSection {
 
 export default function CardSectionComponent({ title, subtitle, cards, className }: CardSection) {
   return (
-    <Section className={classNames(className)}>
+    <Section className={classNames(className, styles.root)}>
       <Container size='large'>
         <div className={styles.cardSection}>
-          <SubHeaderComponent title={title} subtitle={subtitle} />
+          <SubHeaderComponent title={title} subtitle={subtitle} titleWeight='bold' />
```
Lisenish, line 22: Shouldn't we have more semantic prop, maybe _main_ or somehting like that. It just usually a code smell when API just maps 1:1 to CSS props

tan-90: That's a good point. I changed the weights to "primary" and "secondary" since that's mostly how we use them, primary titles being bold and secondary ones being normal.

Lisenish (approved):

PR #82, opened 2021-01-05
SEO page metadata

tan-90:
Added metadata to all pages.
Changed pages to pass metadata to the layout component.
Fixed wrong image path on long-form content.


src/hooks/useSiteMetadata.tsx
```diff
@@ -4,6 +4,7 @@ const defaultMeta = {
   title: 'Default title',
   description: 'Default Description',
   url: 'default',
+  siteUrl: 'default',
```
Lisenish, line 7: Strange, why it was deleted?

Lisenish: BTW, should we refactor this at last? (url & siteUrl duplicates)

src/pages/demo/index.tsx
```diff
@@ -1,26 +1,34 @@
...
+  pageContext: GeneratedPageContext
+}
+export default function DemoPage({ pageContext }: DemoPageProps) {
   const breadcrumbs = pageContext.breadcrumb.crumbs
+  const { pathname } = useLocation()
+  let siteMetadata = useSiteMetadata()
+  siteMetadata = {
+    ...siteMetadata,
+    title: 'Technical Demo - FingerprintJS Pro',
+    url: `${siteMetadata.url}${pathname}`,
```
Lisenish, line 22: Let's use siteUrl instead and replace all places to use it.

src/pages/pricing/index.tsx
```diff
@@ -1,6 +1,5 @@
 import React from 'react'
-import Helmet from 'react-helmet'
-import Layout from '../../components/Layout'
+import { LayoutTemplate } from '../../components/Layout'
```
Lisenish, line 2: I think maybe we should make regular Layout named export as well? So stimulate conscious choice between the two.

tan-90: Agreed!

Lisenish (approved):

Lisenish (approved):

PR #83, opened 2021-01-06
Home ab test

tan-90:
Draft PR for reviewing the home ab test.


src/components/FpjsWidget/index.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useState, useEffect, memo } from 'react'
 import { getVisitTitle, getBrowserName, getBotDecision } from '../../helpers/fpjs-widget'
-import { ReactComponent as InfoSvg } from './info.svg'
+import { ReactComponent as InfoSvg } from '../../img/info.svg'
```
Lisenish, line 3: I think I saw some changes in other PR is it intentional? This branch was created from some other?

Lisenish:
@tan-90 I don't see integration with analytics here, have you discussed how do we plan to analyze results with Savannah? 
Sorry I thaught I attached to the ticket/showed this article before https://www.gatsbyjs.com/docs/how-to/testing/ab-testing-with-google-analytics-and-netlify/ to you, I think we should implement something similar but probably based on GTM dataLayer instead of GA since we're using GTM

src/components/FpjsWidget/index.tsx
Lisenish: @tan-90 ?

Lisenish: Looks like we need to sync develop here

src/helpers/gtm.ts
```diff
@@ -8,7 +10,7 @@ declare global {
 }
 
 export function sendEvent({ event }: { event: string }) {
-  window.dataLayer = window.dataLayer ?? []
+  window.dataLayer = window.dataLayer ?? [{ branch: BRANCH }]
```
Lisenish, line 13: Minor: Let's extract default value to const outside, like const` initialDataLayer =  ... ` and use it here, just for clarity

Lisenish (changes requested):

PR #84, opened 2021-01-07
fix: layout for visitor info icon

spalt08:


src/components/FpjsWidget/FpjsWidget.module.scss
```diff
@@ -206,6 +206,7 @@
           &:not(.browser) {
```
Lisenish, line 206: We also have Mobile widget, let's do the same there, it reproduces only on very small screens there but just in case...

src/components/FpjsWidget/FpjsWidget.module.scss
```diff
@@ -240,9 +245,9 @@
           color: v('brand-fire');
           font-weight: v('bold');
           opacity: 1;
+          margin-left: 0 !important;
```
Lisenish, line 248: I see that there is !important above for the height aldready but it's probably was just migrated from legacy styles, can't we increase specificity to avoid it?

Lisenish (changes requested):

Lisenish:
Still broken if content on the left column is wide enough. 
The thing with this bug it's flaky because it depends on the dates in the left column, and that's why it's not reproduced  for everyone anytime. 
Probably should set mac width and overflow for the left column or forbid grow at all

image

src/components/FpjsWidget/FpjsWidget.module.scss
spalt08: fixed

src/components/FpjsWidget/FpjsWidget.module.scss
spalt08: styles are so complicated, so I am not able to figure out how to remove !important here, sorry
I can spend more time on it later in the end of January

Lisenish: No, it's too extreme for such minor thing to get back to it in Jan, ok it's fine we have a lot of them from legacy one more wouldn't kill anybody, thanks :)

Lisenish (approved):

PR #87, opened 2021-01-08
Unlimited preset for slider and pricing calculator.

tan-90:


src/components/PriceCalculator/index.tsx
```diff
@@ -34,6 +39,24 @@ export default function PriceCalculator() {
...
+
+  function getReservedPrice() {
+    return customCount === undefined
+      ? calculatePrice(selectedPreset.value, PaymentType.Annually)
+      : calculatePrice(
+          customCount >= minimumIdentifications ? customCount : minimumIdentifications,
+          PaymentType.Annually
+        )
+  }
+
```
borislobanov92, line 59: these two methods differ only in payment type, perhaps it would make sense to extract the shared logic and then either use that one method with a parameter, or just call it inside of these methods with the hard coded payment type

src/components/PriceCalculator/index.tsx
```diff
@@ -10,15 +10,20 @@ import { numberFormatter } from '../../helpers/format'
 import styles from './PriceCalculator.module.scss'
 
 export default function PriceCalculator() {
-  const selectOptions = pricingTable.map((entry) => ({
+  const selectOptions = pricingTable.slice(0, 5).map((entry) => ({
     label: numberFormatter.format(entry.value),
     value: entry.value,
   }))
+  selectOptions.push({ label: `${numberFormatter.format(10000000)}+`, value: Infinity })
```
borislobanov92, line 17: why is it different from the Billing page? I'm okay with either logic (although since it's not a dynamic number, I'd just stick to '10M+') but consistency is good

src/components/PriceCalculator/index.tsx
```diff
@@ -10,15 +10,20 @@ import { numberFormatter } from '../../helpers/format'
...
 
   const [selectedPreset, setSelectedPreset] = useState(selectOptions[0])
   const [customCount, setCustomCount] = useState<number | undefined>(undefined)
   const [isContactSalesModalOpen, setIsContactSalesModalOpen] = useState(false)
 
+  function shouldUseSpecialModel() {
```
borislobanov92, line 23: I think that the word 'special' is not descriptive enough. it also doesn't scale very well, what if we have another 'special' model in the future? perhaps `isCustomPricing()` could work a little better?

src/components/Billing/index.tsx
```diff
@@ -14,14 +14,16 @@ import { PATH } from '../../constants/content'
...
 const sliderConfig = {
   min: 0,
-  max: 6,
+  max: 5,
   def: 0,
 }
 
 export default function Billing() {
-  const sliderTable = pricingTable.map(({ label, value }) => {
+  const sliderTable = pricingTable.slice(0, 5).map(({ label, value }) => {
```
borislobanov92, line 22: are we using the pricing table without slicing somewhere else? if not, then why not hardcode the Infinity option into the table right away?

src/components/Billing/index.tsx
```diff
@@ -82,29 +89,53 @@ export default function Billing() {
             <div className={styles.payment}>
               <div>
                 <span className={styles.price}>{monthlyPayment} </span>
-                per month
+                {sliderValue !== sliderConfig.max && 'per month'}
```
borislobanov92, line 92: Do you think it would be better to call the field `monthlyPaymentLabel` and fo this logic in the `recalculatePricing()` method? 'Custom pricing' doesn't sound like a value that fits well into a `monthlyPayment` variable

borislobanov92 (approved):
All comments are minor, so approved in advance

src/components/PriceCalculator/index.tsx
tan-90: I should have noticed that, my bad. Made it a parameter.

src/components/PriceCalculator/index.tsx
tan-90: The select uses the full values, so I thought `10M+` would look out of place:

image

src/components/Billing/index.tsx
tan-90: I removed the 10M and 20M options so we no longer need the slice. But if we add a `{label: '10M+', value: Infinity}` entry, we'll have to stick to the 10M+ label on the slider. I think it looks bad, but it's an easy change, let me know what you think:
image

src/components/Billing/index.tsx
tan-90: Agree, I renamed it to `monthlyPaymentLabel`. I don't know what you meant by "and fo this logic in the recalculatePricing() method?" though.

borislobanov92: crap, part of my message got erased. nevermind, I think it's ok

src/components/Billing/index.tsx
borislobanov92: sure, 10M+ looks out of place

PR #89, opened 2021-01-12
Added context aware env variables

tan-90:
Ability to set environment variables based on netlify context.
Used to set FPJS_TOKEN to a different value when running on deploy preview.

Closes WEB-89


src/constants/env.ts
```diff
@@ -1,4 +1,17 @@
...
+  BranchDeploy,
+}
+const context: NetlifyContext = getNetlifyContext()
+
+export const FPJS_TOKEN =
+  getContextEnv({
+    [NetlifyContext.Production]: process.env.GATSBY_FPJS_TOKEN,
+    [NetlifyContext.DeployPreview]: process.env.GATSBY_PREVIEW_FPJS_TOKEN,
+  }) ?? 'test_client_token'
 export const FPJS_API_TOKEN = process.env.GATSBY_FPJS_API_TOKEN ?? 'test_fpjs_api_token'
```
Lisenish, line 15: Could you please add ability to override these vars also: `FPJS_API_TOKEN`, `FPJS_ENDPOINT`,

Lisenish (changes requested):

src/constants/env.ts
```diff
@@ -1,6 +1,27 @@
...
+  }) ?? 'test_client_token'
+export const FPJS_API_TOKEN =
+  getContextEnv({
+    [NetlifyContext.Production]: process.env.GATSBY_FPJS_API_TOKEN,
+    [NetlifyContext.DeployPreview]: process.env.GATSBY_PREVIEW_API_TOKEN,
+  }) ?? 'test_fpjs_api_token'
+export const FPJS_ENDPOINT =
+  getContextEnv({
+    [NetlifyContext.Production]: process.env.GATSBY_FPJS_ENDPOINT,
+    [NetlifyContext.DeployPreview]: process.env.GATSBY_PREVIEW_API_TOKEN,
```
Lisenish, line 23: I see an error here

Lisenish: I set it as `GATSBY_PREVIEW_FPJS_ENDPOINT` in Netlify so if you fix it should work on DP, please check

Lisenish:
@tan-90 Looks like values is not overwritten on DP. Have you checked it on local env with 'deploy-preview' context value?

Lisenish (approved):

Lisenish (approved):

PR #91, opened 2021-01-12
Ignore css ordering warnings

tan-90:
When using SCSS modules, it's safe to ignore ordering warnings: https://stackoverflow.com/questions/63124432/how-do-i-configure-mini-css-extract-plugin-in-gatsby

Closes WEB-82


gatsby-node.js
```diff
@@ -164,3 +164,14 @@ exports.onCreateNode = ({ node, actions, getNode }) => {
     })
   }
 }
+
+exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
+  if (stage === 'build-javascript') {
+    const config = getConfig()
+    const miniCssExtractPlugin = config.plugins.find((plugin) => plugin.constructor.name === 'MiniCssExtractPlugin')
```
Lisenish, line 171: Can we change it to something like this for additional readability:
```
    const config = getConfig()
    const miniCssExtractPlugin = ignoreCssImportOrder(config)
    actions.replaceWebpackConfig(config)
```
(extract that part to function)

Lisenish (approved):
Approved in advance

Lisenish (approved):

PR #94, opened 2021-01-14
Created case study page layout.

tan-90:
Closes WEB-88


src/components/TitledParagraph/TitledParagraph.module.scss
```diff
@@ -0,0 +1,32 @@
...
+.icon {
+  margin-right: 1.5rem;
+  width: 3rem;
+  height: 3rem;
+  opacity: v('extra-faded');
+  @include transition((opacity));
+}
+
+.title {
+  font-size: 2.75rem;
```
Lisenish, line 30: I see that we often put font styles in-place. Looks like if we decide to change uniform sizes of our headers it would be a lot of work.

We need some good approach for typography, I will create task for this

src/components/TitledParagraph/TitledParagraph.module.scss
```diff
@@ -0,0 +1,32 @@
+.root {
+  margin-top: 4rem;
+
+  & + .root {
```
Lisenish, line 4: Root inside root? Looks strange

src/components/CustomerOverview/CustomerOverview.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import styles from './CustomerOverview.module.scss'
+
+export interface CustomerOverviewProps {
+  logo: string
+  description: string
+  bullets: Array<{ value: string; description: string }>
```
Lisenish, line 8: Probably it's better to extract to own type, when you also can just use `Bullet[]` type

src/pages/case-study/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+import CustomerOverview from '../../components/CustomerOverview/CustomerOverview'
+import TitledParagraph from '../../components/TitledParagraph/TitledParagraph'
+import { Content } from '../../components/Content/Content'
+
+import styles from './case-study.module.scss'
+import Button from '../../components/common/Button'
+
+export default function CaseStudyPage() {
+  const { pathname } = useLocation()
+  let siteMetadata = useSiteMetadata()
```
Lisenish, line 24: I know it's not directly related to this MR but can we refactor this a bit, so it should work like this:

`const siteMetadata = useSiteMetadata({pageOverridesHere})`

src/pages/case-study/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+    },
+    {
+      name: 'Alexa ranking for global internet engagement - Jan. 7 2020',
+      href: 'https://www.alexa.com/siteinfo/chegg.com',
+    },
+    { name: 'Jeffries Virtual Consumer Conference - Jun. 24 2020' },
+  ]
+
+  return (
+    <footer>
```
Lisenish, line 250: It's great that we use semantic html, but we should follow its rules then. 

There can only be one header and footer element within one section, now we have two adjacent footer tags.
You should either change this to section or wrap everything in content section with it's own header and footer and everything else in between

image

src/pages/case-study/case-study.module.scss
```diff
@@ -0,0 +1,113 @@
...
+.subHeader {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  background: adjust-color(c('brand-water'), $lightness: +5%);
+  font-size: 2.25rem;
+}
+
+.downloadPdf {
```
Lisenish, line 34: Shouldn't it be another button variant? Or you don't expect such button on other pages so we treat it like one-off in-place customisation?

src/pages/case-study/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+              logo={companyLogo}
+              description='Chegg is an education company that provides a suite of services including digital and physical textbooks, homework solutions, tutoring, and internship placement.'
+              bullets={[
+                { value: '3.7', description: 'Million Students' },
+                { value: '190', description: 'Countries' },
+                { value: '#275', description: 'Global site rank (2)' },
+              ]}
+            />
+            <BlockQuote author='Augie Kennady, Manager of Consumer Operations & Analytics'>
+              FingerPrintJS helped us solve our account sharing problem in a manner that was both datadriven and, most
```
Lisenish, line 134: Please check `datadriven` spelling with Savannah, I feel like it should be `data-driven`

src/pages/case-study/index.tsx
```diff
@@ -0,0 +1,284 @@
...
+
+      <InlineCtaComponent
+        title='Get in Touch'
+        subtitle='Learn how FingerprintJS Pro can help your business build a custom solution to prevent account sharing and increase revenue.'
+        primaryAction='Get Started Today'
+        buttonText='Start 10-Day Trial'
+        buttonHref='/'
+        secondaryAction='Contact Sales'
+        secondaryLink='sales@fingerprintjs.com'
+        secondaryHref='mailto:sales@fingerprintjs.com'
```
Lisenish, line 261: I think we should replace `mailto` with contact sales popup, but pls discuss with Savannah first

src/pages/case-study/case-study.module.scss
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.problemQuote {
+  margin-bottom: 6rem;
+}
+
+.body {
+  padding: 0 6rem;
+
+  // This will be generated from markdown when we use the CMS so we have to use element selectors.
```
Lisenish, line 89: Shouldn't this styles be used from Content and Blockquote components?

src/components/widgets/SubHeader/index.tsx
```diff
@@ -23,13 +25,48 @@ export default function SubHeaderComponent({
...
+  subtitleWeight = 'secondary',
   align = 'center',
+  labeled,
   className,
 }: SubHeader) {
   return (
     <header className={classNames(styles.root, className, alignmentClasses(align))}>
-      <h1 className={classNames(styles.title, sizeClasses(titleSize), weightClasses(titleWeight))}>{title}</h1>
-      {subtitle && <h3 className={classNames(styles.subtitle, sizeClasses(subtitleSize))}>{subtitle}</h3>}
+      {labeled ? (
```
Lisenish, line 35: Very hard to read and also looks like there's a lot of duplication.

Remember that you can assign elements to variables and use them later
Labeled prop is not clear for me, as I understand it influences subtitle place? Maybe we should call it subtitlePosition = 'top' | 'bottom'? 
Another variant we can have two props `subtitle` and `label` and choose position based on it, but we should rename size and weight props too then

Lisenish (changes requested):

src/components/BlockQuote/BlockQuote.module.scss
```diff
@@ -0,0 +1,46 @@
+.root {
+  display: flex;
+  flex-direction: column;
+
+  position: relative;
```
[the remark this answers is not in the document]
tan-90: The quotes `:after` element doesn't show up otherwise.

src/components/Content/Content.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import classNames from 'classnames'
+
+import styles from './Content.module.scss'
+
+export function HtmlContent({ content, className }: { content: string; className?: string }) {
```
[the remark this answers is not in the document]
tan-90: The `HtmlContent` and `Content` components need to have the same signature so we can use them interchangeably on the CMS to handle both preview and actual rendering. So changing the prop would require the safe component to have the same name. I changed the component name to `DangerouslyRenderHtmlContent` instead.

src/components/TitledParagraph/TitledParagraph.module.scss
tan-90: `& + .root` makes the top margin a bit bigger when there are to `TitledParagraph` (roots) right next to each other.

src/components/CustomerOverview/CustomerOverview.module.scss
```diff
@@ -0,0 +1,45 @@
...
+  margin-bottom: 4rem;
+}
+
+.description {
+  font-size: 2.25rem;
+  margin-bottom: 3rem;
+}
+
+.bullet {
+  display: grid;
```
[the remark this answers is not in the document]
tan-90: The reason for using grid was that we need the gaps to be aligned. With flex that would require 2 extra classes for the number and the label to set flex-grow, which felt a lot less clean.
image

src/pages/case-study/index.tsx
tan-90: The main problem with that is the `siteUrl`. When we use the hooks and override the `siteUrl`, we need to do `${metadata.siteUrl}${pathname}`, otherwise we can't use relative paths.

It's possible to assume the passed override is always relative and do `${metadata.siteUrl}${pathname}` inside the hook, but that feels misleading as override implies the value would be replaced.

Open to discussion on how to handle it.

src/components/Content/Content.module.scss
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+
+  a {
+    text-decoration: none;
+    border-bottom: 1px solid v('brand-fire');
+    color: v('brand-fire');
+  }
+
+  blockquote {
```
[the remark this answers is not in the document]
tan-90: `@use` would be better, but since `@use` statements have to be before every other statement, we wouldn't be able to add them to a module. Having to use the CSS loader config to prepend `@use` directives for the styling of a common component sounds overkill and out of place, so I went with `@import`.

src/pages/case-study/case-study.module.scss
tan-90: The content component allows for a class to be given as an override for the styles of generated content. I used that to make some changes to the default styling as it looked wrong when used to recreate the case-study pdf.

Lisenish (approved):
Approved in advance, please merge after the fixes

src/pages/case-studies/edtech/business-and-finance.svg
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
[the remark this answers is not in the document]
tan-90: Optimized all of them for consistency.

src/templates/case-studies.tsx
```diff
@@ -0,0 +1,87 @@
...
+import React from 'react'
+import Container from '../components/common/Container'
+import Section from '../components/common/Section'
+import { LayoutTemplate } from '../components/Layout'
+import { GeneratedPageContext } from '../helpers/types'
+import { useLocation } from '@reach/router'
+import useSiteMetadata from '../hooks/useSiteMetadata'
+import PaginationNav from '../components/PaginationNav/PaginationNav'
+import PostGrid from '../components/PostGrid/PostGrid'
+// import { PostProps } from '../components/Post/Post'
```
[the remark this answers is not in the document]
tan-90: Yeah, I was writing at least the base for the CMS code, then commenting out some parts and replacing them with hardcoded values. I was thinking about how to handle this before the merge, probably gonna delete all the comments and them get them back from history when we integrate the CMS.

src/components/FeatureList/FeatureList.tsx
```diff
@@ -0,0 +1,33 @@
...
+          </li>
+        ))}
+      </ul>
+    </div>
+  )
+}
+
+function FeatureCheckbox() {
+  return (
+    <span className={styles.checkbox}>
```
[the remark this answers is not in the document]
tan-90: Correct, not sure why I did that. Changed it to be just the styled SVG.

Lisenish:
@tan-90 The link in case studies landing pages doesn't work

Lisenish (changes requested):
@tan-90 The link in case studies landing pages doesn't work

src/templates/long-form-content.tsx
```diff
@@ -74,34 +71,56 @@ export const pageQuery = graphql`
...
+
+  const source = createMemorySource(BASE_URL)
+  const history = createHistory(source)
+
   return (
-    <LongFormContentTemplate
-      post={mapToPost({ frontmatter: entry.get('data').toObject() }, true)}
-      body={widgetFor('body') ?? <></>}
-    />
+    <AppProviders>
```
Lisenish, line 115: Maybe we should extract these to something like `CMSPreviewProviders`?
It's minor duplication now, so up to you, I will approve it

Lisenish (approved):

PR #98, opened 2021-01-19
Added utm parameter forwarding.

tan-90:
UTM parameters from the query string are now sent to the GetStarted and ContactSales forms. Also added as a query string to the Start Trial buttons on the header and mobile nav.

This first iteration does not use cookies and will only forward the parameters that are currently in the query string.


src/hooks/useUtmParams.ts
```diff
@@ -0,0 +1,15 @@
+import { useMemo } from 'react'
+import { useQueryParams } from './useQueryParams'
+
+export const useUtmParams = () => {
+  const queryParams = useQueryParams()
+
+  return useMemo(() => {
+    return Object.keys(queryParams)
+      .filter((key) => key.startsWith('utm_') || key === 'referral_url')
```
Lisenish, line 9: Sorry maybe it wasn't clear from the ticket but you should provide `referral_url` here and on Dashboard side from the actual page's referrer if it's present for it, not try to get it from url search params (https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer)

Lisenish (changes requested):

Lisenish:
For some reason doesn't work for me on DP, @tan-90 does it work for you?

image

Lisenish (changes requested):
LGTM!
I've asked Savannah about trailing slashes, so if she approves we can merge it

Lisenish:
She actually just answered that it should be fine, so let's fix the conflict and merge it

Lisenish (approved):

PR #99, opened 2021-01-19
Removed logo swipper and added more logos.

tan-90:
Changed the client section so it always shows all logos.

Closes WEB-91


Lisenish:
Probably title should't break now when we have wider content, looks strange
image

src/components/ClientsSection/index.tsx
```diff
@@ -5,12 +5,26 @@ import SwiperCore, { Pagination } from 'swiper'
 import Section from '../common/Section'
 import Container from '../common/Container'
 import styles from './ClientsSection.module.scss'
-import classNames from 'classnames'
 
 SwiperCore.use([Pagination])
```
Lisenish, line 9: Why is it still here? I think we removed slider, right? Check also if we even need this lib now at all (probably we do for mobile demo swiper, but not sure)

Lisenish (changes requested):

src/components/ClientsSection/index.tsx
tan-90: We still need it for the mobile demo, but yeah this was my bad. Removing the imports.

Lisenish (approved):

PR #101, opened 2021-01-20
Added CMS markdown support.

tan-90:
Card content, image block text and CTA subtitle for the static page can now be edited as markdown.

Closes WEB-91

### DIFF
--- a/content/index/account-sharing.md
+++ b/content/index/account-sharing.md
@@ -16,24 +16,24 @@ hero:
   ctaText: Create Account
   ctaHref: https://dashboard.fingerprintjs.com/signup
 cardSection:
-  title: "Use FingerprintJS for account sharing prevention to:"
-  subtitle: ""
+  title: 'Use FingerprintJS for account sharing prevention to:'
+  subtitle: ''
   cards:
     - icon: /img/uploads/account_fraud.png
       title: Increase revenue for your business
-      content: Account sharers already know the value of your service. By enforcing
+      markdownContent: Account sharers already know the value of your service. By enforcing
         subscription limits via user identification, you can convert "free"
         account-sharing users to paying customers without negatively impacting
         the rest of your users' experience.
     - icon: /img/uploads/payment_processing.png
       title: Increase security for your customers
-      content: Password sharing increases the likelihood of account takeover, and
+      markdownContent: Password sharing increases the likelihood of account takeover, and
         makes it more difficult for your team to identify real risks. By
         ensuring customers keep their accounts to themselves, you can improve
         account security for all users.
 blocks:
   - subheader: Account sharing prevention built for your business
-    content: Our team works with top SaaS providers to implement custom account
+    markdownContent: Our team works with top SaaS providers to implement custom account
       sharing solutions with the customer experience in mind. With our highly
       accurate visitor identifier, we will work with your team to make sure your
       preferred customers' experience is unhindered, and users breaking their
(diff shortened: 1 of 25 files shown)


content/index/account-sharing.md
```diff
@@ -16,24 +16,24 @@ hero:
...
 cardSection:
-  title: "Use FingerprintJS for account sharing prevention to:"
-  subtitle: ""
+  title: 'Use FingerprintJS for account sharing prevention to:'
+  subtitle: ''
   cards:
     - icon: /img/uploads/account_fraud.png
       title: Increase revenue for your business
-      content: Account sharers already know the value of your service. By enforcing
+      markdownContent: Account sharers already know the value of your service. By enforcing
```
Lisenish, line 24: Maybe we should change it to more visible prefix and more robust to accidental name conflict `markdown__`?
I think probably no need to change everything again now, just an idea, maybe for the future?

src/components/widgets/CardGrid/CardGrid.module.scss
```diff
@@ -24,7 +24,36 @@
   .content {
     margin-top: 1.75rem;
 
-    font-size: 1.75rem;
+    p,
```
Lisenish, line 27: Why this is copied in each client component? I assume we need small correction in each place, probably can't avoid it to get the best results? 

Maybe we should at least extract it to scss mixin with some arguments for customization, so you don't have to copy it over in each place

src/components/Post/Post.tsx
```diff
@@ -52,8 +52,43 @@ export default function Post({
   )
 }
 
-export function mapToPost(data: any): PostProps {
+export function mapToPost(data: any, editing?: boolean): PostProps {
   if (!data.frontmatter || !data.frontmatter.metadata) {
+    // We might not have all the data while editing the post on  the CMS, but we still want the preview to display what we have.
```
Lisenish, line 57: Minor: several levels of `if` nesting is usually a code smell. We can leave the check in the begging of function just add there `&& !editing` and just on if below with editing, it would be easier to read

src/templates/static-page-content.tsx
```diff
@@ -159,17 +159,63 @@ export function StaticPageContentTemplate({
...
+  const [remark, setRemark] = useState()
+  const [remarkHTML, setRemarkHTML] = useState()
+
+  useEffect(() => {
+    async function importRemark() {
+      setRemark(await import('remark'))
+      setRemarkHTML(await import('remark-html'))
+    }
+
+    importRemark()
```
Lisenish, line 172: Parsing markdown content and replacing it

This is similar task, see how https://github.com/react-syntax-highlighter/react-syntax-highlighter approach syntax highlighting, I think it's similar.

We don't need HOC component for this, probably it's a stretch since we need this only for CMS preview pages, so let's do all the parsing work in effect, and we may even extract it to custom hook after that, but it's up to you.

I think it's fine if we get a blink of raw markdown on first render, if it bother someone we could implement processing... state later

Lisenish (changes requested):

src/templates/static-page-content.tsx
Lisenish: Another approach: we probably also could use React.Lazy since it's browser-only page, so should work here. And even if we need dynamically imported components with SSR support I think there is a separate lib for it. But it's probably too much here, let's do what I wrote above

src/components/widgets/CardGrid/CardGrid.module.scss
tan-90: Yeah, the font sizes are always different from the blog post, which is the primary use of the Content component. I'll make a mixin.

content/index/account-sharing.md
tan-90: I agree. It's easy to change with a find and replace, so I did it already.

src/templates/static-page-content.tsx
tan-90: Discussed on call. Changed to a cleaner, but still not ideal approach. Waiting on consistent typography so we can revisit this.

Lisenish (approved):

PR #108, opened 2021-01-27
increased visitorId font size 

natapg:
Closes WEB-93

Increased visitor ID font size to decrease distance with info icon


natapg:
With a font size of 2.75rem the visitorId is very close to the logo, I think it looks better with 2.70rem

2.75rem
image

2.70rem
image

Lisenish (approved):

PR #111, opened 2021-01-29
Added light variant for contact sales form.

tan-90:
Makes the messages black so they show up on light backgrounds.

### DIFF
--- a/src/components/ContactSalesForm/ContactSalesForm.module.scss
+++ b/src/components/ContactSalesForm/ContactSalesForm.module.scss
@@ -69,6 +69,10 @@
         margin-left: 0;
       }
     }
+
+    .messageLight {
+      color: c('black');
+    }
   }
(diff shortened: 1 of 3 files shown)


src/components/ContactSalesForm/ContactSalesForm.module.scss
```diff
@@ -69,6 +69,10 @@
         margin-left: 0;
       }
     }
+
+    .messageLight {
```
Lisenish, line 73: For the future, we can follow BEM notation for such things (https://css-tricks.com/bem-101/)

We don't need to use the full blown notation since we use CSS modules, but I think it's reasonable for modifiers.
So it would be `.message--light` in this case

Lisenish (approved):

Lisenish (approved):

PR #113, opened 2021-02-02
Test PR for code checks.

tan-90:
Test PR for the lighthouse app.

(expecting both lint and bundlewatch to fail)


package.json
```diff
@@ -105,6 +114,7 @@
     "lint:fix": "eslint . --ext .ts,.tsx,.js --quiet --fix",
     "test": "echo \"Error: no test specified\" && exit 1",
     "storybook": "start-storybook -p 6006 -s ./static",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "prettier": "prettier -c ."
```
Lisenish, line 118: Maybe we should add fix command shortcut as well?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+on: [push]
+jobs:
+  checks:
+    name: Code checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: gatsby .cache folder
```
Lisenish, line 11: 1. Probably need a more clear name? Like "Cache gatsby .cache folder"?
1. Shouldn't we also cache npm deps ('.npm')? Also see `cache-hit` return result to skip installation step then
1. There is v2 version, shouldn't we use it?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/cache@v1
+        id: gatsby-public-folder
+        with:
+          path: public/
+          key: ${{ runner.os }}-public-gatsby
+          restore-keys: |
+            ${{ runner.os }}-public-gatsby
+
+      - name: Use Node.js 15.x
+        uses: actions/setup-node@v1
```
Lisenish, line 30: Same question about caching itself node installation, but maybe it's integrated into the action itself

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          node-version: 15.x
+
+      - name: yarn install
+        uses: bahmutov/npm-install@v1
```
Lisenish, line 35: Why do you use some action here? Shouldn't it be just simple `run`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+          yarn global add @lhci/cli@0.6.x
+          export PATH=$(yarn global bin):$PATH
+          lhci autorun
+        env:
+          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
+
+      - name: run bundlewatch
+        run: yarn run bundlewatch
+        env:
+          BUNDLEWATCH_GITHUB_TOKEN: ${{ secrets.BUNDLEWATCH_GITHUB_TOKEN }}
```
Lisenish, line 61: Requested  from Valentin to provide an access to this

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: run prettier
+        run: yarn run prettier
+
+      - name: run Lighthouse CI
+        run: |
+          yarn global add @lhci/cli@0.6.x
+          export PATH=$(yarn global bin):$PATH
+          lhci autorun
+        env:
+          LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
```
Lisenish, line 56: Looks like it doesn't work? I don't see status check from LH :(

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
...
+        env:
+          GATSBY_EXPERIMENTAL_PAGE_BUILD_ON_DATA_CHANGES: true
+
+      - name: run eslint
+        run: yarn run lint
+
+      - name: run prettier
+        run: yarn run prettier
+
+      - name: run Lighthouse CI
```
Lisenish, line 50: Shouldn't we also add assertions? 
Our score should be good now
https://github.com/GoogleChrome/lighthouse-ci/blob/master/docs/getting-started.md#add-assertions

.github/workflows/ci.yml
```diff
@@ -0,0 +1,61 @@
+name: CI
+on: [push]
+jobs:
+  checks:
+    name: Code checks
```
Lisenish, line 5: Let's maybe name it CI checks: lint, prettier, build for clarity

src/templates/blog-featured.tsx
```diff
@@ -46,7 +46,7 @@ export default function BlogFeatured({ data, pageContext }: BlogFeaturedProps) {
         <Container size='large'>
           <h1>Featured Articles</h1>
 
-          <PostGrid posts={posts.map(({ node }) => node).map(mapToPost)} />
+          <PostGrid posts={posts.map(({ node }) => node).map((node) => mapToPost(node))} />
```
Lisenish, line 49: What's was the error btw? Different count of arguments?

Lisenish:
@tan-90 Looks like we have warning to fix. 
(also could you please check why it didn't fail the build, because AFAIK github actions should set env var CI=true and in this case all warning should be treated as errors on CI)
image

package.json
tan-90: Added `prettier:fix`. Since `lint:fix` uses the `--quiet` flag, I set the log level to error for this new command as well.

.github/workflows/ci.yml
tan-90: Changed to `actions/setup-node@v2`. This one should [handle caching](https://github.com/actions/setup-node#v2).

.github/workflows/ci.yml
tan-90: 1. Renamed the cache actions.
2. I used an action that handles caching.
3. Changed to v2.

.github/workflows/ci.yml
tan-90: [This action](https://github.com/bahmutov/npm-install#basic) handles caching the installation.

.github/workflows/ci.yml
tan-90: Not sure about that either. I double-checked and `LHCI_GITHUB_APP_TOKEN ` is the correct name, but the log says lighthouse couldn't find the token.

.github/workflows/ci.yml
tan-90: It failed with a lot of errors when I set the assertion to that preset. I don't think github actions lets you see previous logs, but I can add it back if you want to take a look.

src/templates/blog-featured.tsx
tan-90: Yeah, `mapToPost` now also takes an optional `preview` argument which handles client side markdown rendering for the CMS preview, so we can't pass it directly to `map` anymore.

.github/workflows/ci.yml
Lisenish: Done

.github/workflows/ci.yml
Lisenish: @tan-90 Could you please test it? Should be fine now

.github/workflows/ci.yml
Lisenish: That's strange but ok, we can do it gradually, let's start without them first

.github/workflows/ci.yml
Lisenish: Maybe it's not the best for security reasons but I think it's fine here

Lisenish (approved):

PR #116, opened 2021-02-08
 fixed navigation bar on scroll

natapg:
Closes WEB-105

### DIFF
--- a/src/components/Header/Header.module.scss
+++ b/src/components/Header/Header.module.scss
@@ -1,6 +1,13 @@
 .nav {
   padding: 3rem 0 0;
 
+  @include media('>desktop') {
+    position: fixed;
+    width: 100%;
+    top: 44.8px;
+    z-index: 9999;
+    background: white;
+  }
   @include media('<=desktop') {
     margin-top: 0;
     padding-top: 2rem;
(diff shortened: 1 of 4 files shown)


natapg:
I have updated the PR using sticky position instead of fixed position, this way it is not necessary to add margin or padding to the content since the header is still part of the flow of the page @Lisenish

src/components/Header/Header.module.scss
```diff
@@ -1,6 +1,17 @@
+.header {
+  @include media('>desktop') {
+    position: sticky;
+    top: 0;
+    z-index: 9999;
```
tan-90, line 5: `custom-properties.scss` has `z-index` variables, we should use those.

src/components/Header/Header.module.scss
```diff
@@ -1,6 +1,17 @@
...
+  }
+}
+
 .nav {
-  padding: 3rem 0 0;
+  padding: 3rem 0 1rem;
 
+  @include media('>desktop') {
+    background: white;
```
tan-90, line 13: There's a variable for white as well, you can get the value with `c('white')`.

tan-90 (changes requested):

tan-90 (approved):

PR #118, opened 2021-02-12
fix: fixed deploy preview demo

Lisenish:
Added TLS param for fpjs calls, we should use staging TLS endpoint for the deploy preview


tan-90 (approved):
Everything looks good to me.

PR #126, opened 2021-03-10
Draft PR for mini-apps tests

natapg:


content/blog/the-beginner’s-guide-to-browser-fingerprinting-for-fraud-detection.md
```diff
@@ -7,7 +7,7 @@ metadata:
     technique works with a real-life example.
   url: https://fingerprintjs.com/blog/what-is-browser-fingerprinting
   image: /img/uploads/3.png
-featured: true
+featured: false
```
Lisenish, line 10: I think it could be multiple featured articles and it will show the last one as the biggest one, so probably we may leave it as featured as well

src/components/Content/Content.module.scss
```diff
@@ -54,11 +78,14 @@
   }
 
   ul {
-    list-style: '>  ' outside none;
     padding-left: 4rem;
+    list-style-type: disc;
```
Lisenish, line 82: Let's also put a comment why we're using botth `marker` and `list-style-type`

static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: feature/web-118-integrate-mini-apps-blog
```
Lisenish, line 3: Don't forget to change back before the merge

Lisenish (commented):
@natapg Only minor things, great job

I can't comment on binary files, but do we need all of them? Looks like there may be some unnecessary ones.

Lisenish (approved):

PR #131, opened 2021-03-19
chore: updated browserlist config

Lisenish:


tan-90 (approved):
Looks good to me. (glad we're dropping IE)

PR #135, opened 2021-03-24
feature: free-tier Pricing changes

natapg:


src/components/Billing/index.tsx
```diff
@@ -41,13 +41,21 @@ export default function Billing() {
   }
 
   const recalculatePricing = (value: number, paymentType: PaymentType) => {
-    if (value === Infinity) {
-      setMonthlyPaymentLabel('Custom pricing')
-      return
+    switch (value) {
+      case Infinity: {
```
Lisenish, line 45: Minor: You don't need curly braces for cases without new vars

src/components/PriceCalculator/index.tsx
```diff
@@ -14,14 +14,21 @@ export default function PriceCalculator() {
...
   const [selectedPreset, setSelectedPreset] = useState(selectOptions[0])
   const [customCount, setCustomCount] = useState<number | undefined>(undefined)
   const [isContactSalesModalOpen, setIsContactSalesModalOpen] = useState(false)
 
   function isCustomPricing() {
-    return (!customCount && selectedPreset.value === Infinity) || (customCount && customCount >= 10000000)
+    return (!customCount && selectedPreset.value === Infinity) || (customCount && customCount >= 500000)
+  }
+
+  function isFree() {
```
Lisenish, line 27: BTW, I'm not sure why this and isCustomPricing are implemented as function and not as just const variables.

Maybe to preserve some CPU power because if it's isFree we don't need to calculate isCustomPricing, but I think it's useless optimisation

Lisenish (approved):
Only minor things, approved

src/components/Billing/index.tsx
natapg: thanks, as the last case has a variable, I put braces in all cases. Fixed

src/components/PriceCalculator/index.tsx
natapg: I've changed the functions by const variables

src/components/PriceCalculator/index.tsx
```diff
@@ -20,11 +20,11 @@ export default function PriceCalculator() {
   const [customCount, setCustomCount] = useState<number | undefined>(undefined)
   const [isContactSalesModalOpen, setIsContactSalesModalOpen] = useState(false)
 
-  function isCustomPricing() {
+  const isCustomPricing = () => {
```
Lisenish, line 23: I meant that you can just make it a bool variable like `const isCustomPricing = (!customCount && selectedPreset.value === Infinity) || (customCount && customCount >= 500000)`

Because there aren't any params in these functions and they only depends on clousers

natapg: I see now, my bad!

Lisenish:
@natapg I've already approved it, so you could merge it by yourself after the latest fix

PR #136, opened 2021-03-24
Free Tier release updates

natapg:


Lisenish (commented):
Looks good, but we shouldn't merge it until the free-tier release

static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: release/free-tier
```
Lisenish, line 3: Don't forget to change back (I think Savannah finished with all required updates?

natapg: I'll change it before merging to developer, just in case we find something else to change

natapg: changed back

Lisenish (approved):

PR #142, opened 2021-03-30
fix: problem with images that don't render well in blog articles 

natapg:
Closes WEB-121
removed 2nd transformer remark plugin that was causing the problem

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -91,17 +91,10 @@ module.exports = {
             },
           },
           {
-            resolve: `gatsby-transformer-remark`,
+            resolve: `gatsby-remark-highlight-code`,
             options: {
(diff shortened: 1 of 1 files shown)


gatsby-config.js
```diff
@@ -91,17 +91,10 @@ module.exports = {
             },
           },
           {
-            resolve: `gatsby-transformer-remark`,
+            resolve: `gatsby-remark-highlight-code`,
```
Lisenish, line 94: @natapg Shouldn't it be inside plugins of 'gatsby-transformer-remark'? 
Now you've added it just as separate Gatsby plugin, you said it works but looks suspicous.

Lisenish (approved):

PR #152, opened 2021-04-09
feature: Feature Comparison section

natapg:
ticket : fpjs.myjetbrains.com/youtrack/issue/WEB-115
Mock-Up: http://pro.fingerprintjs.com/github


src/components/githubPage/WhyGoProSection/WhyGoProSection.module.scss
```diff
@@ -0,0 +1,64 @@
+.root {
+  background-color: #161e58;
```
tan-90, line 2: Could this be reused somewhere else? If so maybe it could be a variable.

tan-90 (approved):
Looks good to me.

src/components/githubPage/WhyGoProSection/WhyGoProSection.module.scss
natapg: It is used in another component too, I am going to change it to a variable

natapg: I changed to use a modified version of the color brand-water since it is the base color.
I also added some conditions to make the table look good in the mobile view

PR #154, opened 2021-04-12
Adapted tracking events to free tier

tan-90:


src/helpers/gtm.ts
```diff
@@ -9,9 +9,59 @@ declare global {
...
 }
+
+export function sendLegacyEvent(props: SendEventProps) {
+  // Required for development since without SSR sendEvent will be called before Helmet has a chance to inject the script that initializes dataLayer.
+  window.dataLayer = window.dataLayer ?? defaultDataLayer
+
+  window.dataLayer.push({ event: EventType.LegacyEvent, eventProps: { ...props } })
+}
+
+export function trackEmbeddedFormClick() {
```
Lisenish, line 53: These event's looks like they could be just set upped in GTM itself without code changes (because they don't depend on API answers)

And we probably shouldn't have two methods of sending events anyway. We shoyld migrate old approach to new if we need a new one. 

Let's discuss this one on the call

Btw it's called `sendLegacyEvent` in Dashboard because we supported old events when we had rewritten our old MVP Dashboard with KD, and we supported old non GTM events, here it's not legacy at all :)
(and I actually think we should rename it in Dash as well since we end up using this event in for new events as well)

Lisenish (changes requested):

src/helpers/gtm.ts
```diff
@@ -9,9 +9,46 @@ declare global {
...
   window.dataLayer = window.dataLayer ?? defaultDataLayer
+}
+
+export function sendGaEvent({ event }: { event: string }) {
+  setupDataLayer()
 
   window.dataLayer.push({ event })
 }
+
+export function sendEvent(props: SendEventProps) {
```
Lisenish, line 46: Sorry for the confusion, but I think this one should be `sendGAEvent`, because it's proxied to GA via legacyEvent in GTM.

The one we use for `optimize.activate` should be called just `sendEvent` because it's for any custom GTM event

Lisenish (changes requested):

src/components/ContactSalesForm/index.tsx
```diff
@@ -34,7 +34,7 @@ export default function ContactSalesForm({ variant = 'dark', className }: Contac
       setTimeout(() => {
         updateFormState(FormState.Default)
       }, 2500)
-      sendEvent({ event: 'leadSubmit.error' })
+      sendGaEvent({ event: 'leadSubmit.error' })
```
Lisenish, line 37: Let's migrate this events to events legacyEvent approach for uniformity, you can take it's GA params from GTM tags

src/helpers/gtm.ts
Lisenish: Actually let's just remove generic sendEvent for now, we have a task to delete Optimise snippet from the head, then we can delete `optimize.activate` event and migrate other two to this approach as well

Lisenish (changes requested):

Lisenish (approved):

PR #162, opened 2021-04-19
[DRAFT] DP for Savvanah to test blog articles changes

natapg:
Savannah needs a DP to test the changes to the new articles.


static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: draft/draft-PR-for-Savannah
```
tan-90, line 3: Don't forget to change back

tan-90 (approved):

static/admin/config.yml
natapg: ok! changed back

PR #168, opened 2021-04-23
fix : removed Google Optimize snippet and changed info icon color

natapg:
Closes WEB-130


tan-90 (approved):
Looks good to me. Didn't think the green would work close to the button but the icon is thin enough it works well as a highlight.

PR #170, opened 2021-04-26
fix images cut out issue

natapg:


tan-90 (approved):
This fix works for my screen size and works for different screen sizes tested using responsive tools.

PR #173, opened 2021-05-06
feat: added enable tracking switch to gtm send event

tan-90:
Paired up with some GTM changes. This makes it so `sendEvent` also sends an `enableTracking` boolean, which is used on GTM to filter which events can trigger.

Closes WEB-137

### DIFF
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -30,7 +30,7 @@ export function LayoutTemplate({ children, siteMetadata }: LayoutTemplateProps) 
   const fpjsEndpoint = FPJS_ENDPOINT
   const gtmToken = GTM_TOKEN
   const { visitorData } = useVisitorData()
-  const shouldEnableAnalytics = visitorData && visitorData.ipLocation.continent?.code?.toUpperCase() === 'EU'
+  const shouldEnableAnalytics = visitorData && visitorData.ipLocation.continent?.code?.toUpperCase() !== 'EU'
 
   useEffect(() => {
     if (shouldEnableAnalytics) {
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { Helmet } from 'react-helmet'
 import Footer from '../Footer'
 import Header from '../Header'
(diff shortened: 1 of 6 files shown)


src/helpers/gtm.ts
```diff
@@ -41,20 +42,34 @@ function setupDataLayer() {
...
+const trackLeadSubmit = (enableTracking: boolean) => (success = true) => {
+  sendEvent(enableTracking, {
     action: EventAction.LeadSubmit,
     category: EventCategory.Lead,
     label: success ? EventLabel.Success : EventLabel.Error,
   })
 }
+
+export function useGtm() {
+  const { visitorData } = useVisitorData()
```
Lisenish, line 68: `visitorData` will be cached for the next renders and pages, right?

src/helpers/gtm.ts
```diff
@@ -41,20 +42,34 @@ function setupDataLayer() {
...
+  const enableTracking = !visitorData || visitorData.ipLocation.continent?.code === 'EU'
+
+  return {
```
Lisenish, line 71: I think this approach with factory is too complex and requires a lot of unnecessary code. 

My suggestion was just to create a special custom event in case we want to enable tracking (not the data layer variable).
And insert GA snippet only after this event is triggered

Lisenish (changes requested):

src/helpers/gtm.ts
```diff
@@ -41,20 +42,34 @@ function setupDataLayer() {
...
+  sendEvent(enableTracking, {
     action: EventAction.LeadSubmit,
     category: EventCategory.Lead,
     label: success ? EventLabel.Success : EventLabel.Error,
   })
 }
+
+export function useGtm() {
+  const { visitorData } = useVisitorData()
+  const enableTracking = !visitorData || visitorData.ipLocation.continent?.code === 'EU'
```
Lisenish, line 69: To avoid any surprises with letters capitalisation we can do` ?.toUpperCase()` here

src/components/Layout/index.tsx
```diff
@@ -27,6 +29,14 @@ export function LayoutTemplate({ children, siteMetadata }: LayoutTemplateProps)
   const { title, description, siteUrl, image } = siteMetadata
   const fpjsEndpoint = FPJS_ENDPOINT
   const gtmToken = GTM_TOKEN
+  const { visitorData } = useVisitorData()
+  const shouldEnableAnalytics = visitorData && visitorData.ipLocation.continent?.code?.toUpperCase() === 'EU'
```
Lisenish: Looks like you forgot to change `=== 'EU'` condition

Lisenish (approved):

PR #177, opened 2021-05-11
Blog authors

tan-90:
Closes WEB-151


gatsby-node.js
```diff
@@ -73,7 +74,7 @@ function getRelativeUrl(url) {
   return relativeUrl ? withTrailingSlash(relativeUrl[1]) : '/'
 }
 
-function createPageFromEdge(edge, createPage) {
+function createPageFromEdge(edge, createPage, additionalContext = {}) {
```
Lisenish, line 77: Do we still need this additionalContext?

gatsby-node.js
```diff
@@ -211,3 +213,28 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+  const { createNodeField } = actions
+
+  const blogPosts = getNodes().filter(
+    (node) => node.internal.type === 'MarkdownRemark' && /(blog)\/.*.md$/.test(node.fileAbsolutePath)
+  )
+  const authors = getNodes().filter(
+    (node) => node.internal.type === 'MarkdownRemark' && /(author)\/.*.md$/.test(node.fileAbsolutePath)
+  )
+
+  blogPosts.forEach((node) => {
```
Lisenish, line 227: I thought we don't need it anymore, forgot to delete?

src/components/common/PreviewCompatibleImage/PreviewCompatibleImage.tsx
```diff
@@ -45,11 +46,10 @@ const PreviewCompatibleImage = ({ className, imageInfo }: Props) => {
   }
 
   if (childImageSharp && isFixed(childImageSharp)) {
-    return <Img className={className} style={imageStyle} fixed={childImageSharp.fixed} alt={alt} />
+    return <Img className={className} style={style} fixed={childImageSharp.fixed} alt={alt} />
   }
 
-  if (!!image && typeof image === 'string')
-    return <img className={className} style={imageStyle} src={image} alt={alt} />
+  if (!!image && typeof image === 'string') return <img className={className} style={style} src={image} alt={alt} />
```
Lisenish, line 52: Minor: let's add curly braces here

src/components/Author/Author.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export interface AuthorComponentProps {
+  author: Author
+  className?: string
+}
+
+export default function AuthorComponent({ author, className }: AuthorComponentProps) {
+  return (
+    <div className={classNames(className, styles.root)}>
+      <PreviewCompatibleImage imageInfo={author.photo} className={styles.photo} imageStyle={{ borderRadius: '6rem' }} />
```
Lisenish, line 22: Why you inlined borderRadius style and not just put it in css?

Lisenish:
There is uploaded image probalby we can delete it before the merge

Lisenish (approved):

src/components/Author/Author.tsx
tan-90: The `style` attribute takes precedence so the CSS would never be applied.

gatsby-node.js
tan-90: Not anymore, but I left it there in case we need it for the future. Could easily revert it though.

gatsby-node.js
tan-90: The gatsby mapping itself is not enough, we still need to create the field. The difference is that now we only do it once and make it a part of the schema, instead of only doing it on page creation and only passing it as context.

src/components/common/PreviewCompatibleImage/PreviewCompatibleImage.tsx
tan-90: [This](https://github.com/fingerprintjs/fingerprintjs.com/pull/177/commits/102574769ca60313ae3db7113fe6df9f37d9697c) is the commit where I removed the code for the old method.

gatsby-node.js
tan-90: [This](https://github.com/fingerprintjs/fingerprintjs.com/pull/177/commits/102574769ca60313ae3db7113fe6df9f37d9697c) is the commit where I removed the code for the old method.

PR #178, opened 2021-05-11
Free trial copy update

natapg:
Closes WEB-145

### DIFF
--- a/src/components/GetStartedForm/index.tsx
+++ b/src/components/GetStartedForm/index.tsx
@@ -107,7 +107,7 @@ export default function GetStartedForm({ className }: GetStartedFormProps) {
           <CheckSvg className={styles.check} />
           <div>No Credit Card Required</div>
         </li>
-        <li className={styles.treeColumns}>
+        <li className={styles.threeColumns}>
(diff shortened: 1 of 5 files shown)


src/components/GetStartedForm/index.tsx
```diff
@@ -101,15 +101,15 @@ export default function GetStartedForm({ className }: GetStartedFormProps) {
       <ul className={styles.description}>
         <li>
           <CheckSvg className={styles.check} />
-          1K Visitors/Month Free
+          <div>Unlimited Identifications for 30 Days</div>
```
tan-90, line 104: I assume we need the divs to limit the width?

tan-90 (approved):
Looks good to me, only one minor question.

src/components/GetStartedForm/index.tsx
natapg: I added them to delimit the width of each column, but maybe not necessary, let me try

natapg: You're right, no need to add the divs.
I added them at the beginning because I needed them but then I used another approach, thanks @tan-90

PR #180, opened 2021-05-13
Changed author style to match figma

tan-90:


src/templates/long-form-content.module.scss
```diff
@@ -26,22 +26,22 @@
...
+  @include media('>1250px') {
     position: absolute;
     flex-direction: column;
 
     margin-top: 16rem;
 
-    // Container padding - user component width
-    left: calc(50vw - 680px / 2 - 320px);
+    // Container padding to match header
+    left: unquote('max(3rem, 50vw - 1232px / 2)');
```
Lisenish, line 36: Maybe we should use a variable for this? Or it's impossilbe? 
But a proper way to do this would be probably to refactor layout to use grid and 3 columns for this, so we don't depend directly on the some padding value

Lisenish (approved):

src/templates/long-form-content.module.scss
tan-90: Maybe we could have a mixin for calculating the container padding. But I do agree that a permanent solution would probably be to refactor the layout and use 3 columns.

PR #182, opened 2021-05-13
fix: reduced queried image size for user avatar

tan-90:


Lisenish (approved):

src/templates/long-form-content.tsx
```diff
@@ -63,7 +63,7 @@ export const pageQuery = graphql`
             role
             photo {
               childImageSharp {
-                fluid(maxWidth: 2048, quality: 100) {
+                fluid(maxWidth: 192, quality: 100) {
```
Lisenish, line 66: Btw, wouldn't it better just use fixed? Since I think it's actually fixed size image

Lisenish (approved):

PR #183, opened 2021-05-14
feat: added banner to pricing page

natapg:
Closes WEB-149


tan-90 (approved):
Looks good to me.

src/pages/pricing/Pricing.module.scss
```diff
@@ -195,3 +195,7 @@
     }
   }
 }
+
+.bannerSection {
+  padding: 5rem 0;
```
Lisenish, line 200: I think we should start avoiding using rem's for anything but typography since it would be pretty hard to do everything pixel perfect according to the mocks.

Lisenish: @tan-90 @natapg FYI

Lisenish (approved):
@natapg Nice job 👍

src/pages/pricing/Pricing.module.scss
tan-90: I use this: https://nekocalc.com/px-to-rem-converter. Our base font size is 8px.

natapg: You mean to measure the distance in pixels in the mock and use that value instead of rems to be more precise?

natapg: image

natapg: thanks for the page @tan-90 , in this case there are 19rems = 152px quite close but not the same

natapg: fixed

Lisenish: @natapg Thanks, I wanted to discuss it in general also didn't even mean.
Btw

@tan-90 Yeah I actually wanted to avoid  jumping through such hoop. But I googled a bit an there're actually pretyy obvious and simple solution:
```
@function rem($size) {
  @return $size / 16px * 1rem;
}
```

@tan-90 @natapg we may use this function and still provide px values but we will preserve accessibility advantages of `rem`. WDYT?

tan-90: > You mean to measure the distance in pixels in the mock and use that value instead of rems to be more precise?

When using REMs in general, even if we move away from it, it will still be quite useful for typography.

tan-90: > @natapg Thanks, I wanted to discuss it in general also didn't even mean.
> Btw
> 
> @tan-90 Yeah I actually wanted to avoid jumping through such hoops. But I googled a bit an there're actually pretyy obvious and simple solution for SCSS:
> 
> ```
> @function rem($size) {
>   @return $size / 16px * 1rem;
> }
> ```
> 
> @tan-90 @natapg we may use this function and still provide px values but we will preserve accessibility advantages of `rem`. WDYT?
> 
> (Need to change base size to 8px of course, btw we have pretty low base for font size, strange )

I like this, would be very practical. The tool helps, but doing a conversion every time takes a bit of time, so going with this could be the best of both worlds.

tan-90: I think the small font size is a leftover from before the migration to gatsby. I remember the text being a lot smaller at first and then we adjusted the typography.

Lisenish: @natapg @tan-90 Ok let's add this function in this PR and start using it and we can merge it already, we have even 2 approvals here

natapg: @Lisenish @tan-90 sorry guys I was out, I've made the changes (added a functions scss file with the function)

@tan-90 I can't merge it (I'm not authorized) if the changes are ok please do it :)

tan-90: I'm also not authorized. @Lisenish

PR #184, opened 2021-05-17
fix: fix google console redirect error report

Lisenish:
Google reported in Search Console that we have incorrect redirect on http://www.fingerprintjs.com


natapg (approved):
LGTM

tan-90 (approved):

PR #187, opened 2021-05-24
Case study article Promotion Abuse

natapg:
Closes WEB-156


lighthouserc.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
         'http://localhost/demo/index.html',
         'http://localhost/pricing/index.html',
         'http://localhost/case-studies/edtech/index.html',
+        'http://localhost/case-studies/promo-abuse/index.html',
```
tan-90, line 15: I don't think we need to add all case studies to lighthouse since they're all gonna be very similar.

src/pages/case-studies/promo-abuse/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+      </Modal>
+    </>
+  )
+}
+
+interface ReferenceListProps {
+  references: Array<{ name: string; href?: string }>
+}
+// TODO [VL] Change this to a widget when integrating the page with the CMS.
+export function ReferenceList({ references }: ReferenceListProps) {
```
tan-90, line 270: We don't need to duplicate this here since it's already exported from the other case study. Now that I think about it it's not used there either, but I guess it stayed as a leftover until we turn it into a widget.

tan-90 (approved):
Just 2 minor comments, approving in advance.

lighthouserc.js
natapg: yes, you're right, I'm going to remove it

PR #189, opened 2021-05-26
Add header bar with link to Venturebeat article

natapg:
Closes WEB-136

Mockups
https://www.figma.com/file/Hdhz29N50l6uql6r7e1iWg/FP-Design-System?node-id=13%3A56 in 'Header Bar' page


src/components/Header/index.tsx
```diff
@@ -46,6 +51,8 @@ export default function Header() {
   return (
     <>
       <header className={styles.header}>
+        {headerBarTitle && headerBarHref && <HeaderBar href={headerBarHref}>{headerBarTitle}</HeaderBar>}
+        {headerBarTitle && !headerBarHref && <HeaderBar>{headerBarTitle}</HeaderBar>}
```
tan-90, line 55: Since  `href` is optional can't we just check for the title?

```suggestion
        {headerBarTitle && <HeaderBar href={headerBarHref}>{headerBarTitle}</HeaderBar>}
```

tan-90 (approved):
Minor comment, approving in advance.

src/components/Header/index.tsx
natapg: you are right, thanks! I will change it

PR #193, opened 2021-05-31
Added disqus

tan-90:
Closes WEB-132


src/templates/long-form-content.tsx
```diff
@@ -97,6 +106,7 @@ export interface TemplateProps {
   breadcrumbs?: Array<Breadcrumb>
   authors?: Author[]
   publishDate?: string
+  disqusConfig?: { url: string; identifier: string; title: string }
```
tan-90, line 109: Considered creating an interface for this but it didn't seem necessary since it's only gonna be used here.

src/templates/long-form-content.tsx
```diff
@@ -137,6 +148,13 @@ export function LongFormContentTemplate({
         <Container>
           <RelatedArticles article={post} />
         </Container>
+
+        {disqusConfig && (
+          <Container className={styles.comments}>
+            <h2>Discussion</h2>
+            <Disqus config={disqusConfig} />
+          </Container>
+        )}
```
tan-90, line 157: Conditional rendering hides it on the CMS preview since there we wouldn't have the slug and full metadata.

Lisenish:
@tan-90 This may conflict with our CSP changes (https://github.com/fingerprintjs/fingerprintjs.com/pull/196). After CSP will be added we need to retest it and add all necessary values to CSP (if it doesn't generate them automatically)

natapg (approved):
LGTM 👍🏻 👍🏻

Lisenish (approved):
LGTM 👍 
@tan-90 let's merge but delete the comments first

PR #196, opened 2021-06-03
fix: another CSP iteration

Lisenish:
Due to a lot dynamic inline scripts from Gatsby build we can't just hardcode the hashes, so I had to use this plugin to generate them during build time.
Downside: it uses meta tag instead of generating headers file for Netlify, so we can't use report-uri here and some other attributes. Theoretically we could adapt it so it generates headers file an merges it's content but I think it would be an overhead at this point

Closes WEB-163

### DIFF
--- a/config/csp-config.js
+++ b/config/csp-config.js
@@ -1,0 +1,38 @@
+const isProdContext = process.env.CONTEXT === 'production' // can't use env helper from TS sources here
+
+const cspProdValues = {
+  scriptSrc:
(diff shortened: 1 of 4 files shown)


borislobanov92 (approved):

tan-90 (approved):

Lisenish:
Discussed with Valentin and it seems there is more difficulties and inconveniences than good from CSP on the website so we decided to remove it.

So I'm closing this

PR #203, opened 2021-06-14
added new case study article - third party verification

natapg:
Closes WEB-165


tan-90 (approved):
Looks good to me.

With 3 case studies already we might wanna start thinking about adding support to create new ones from the CMS.

PR #210, opened 2021-06-25
fix: copy changes for API call pricing

natapg:
Closes WEB-172


static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: fix/WEB-172-pricing-experiment
```
natapg, line 3: just in case Savannah wants to make a change

tan-90 (approved):

natapg:
@tan-90 @Lisenish I've updated the blog articles and added a filter so that the article that advertises the free-tier is not displayed in the article list (remains accessible by URL)

src/helpers/pricing.ts
```diff
@@ -28,10 +28,10 @@ export function calculatePrice(price: number, type: PaymentType): string {
   }
 
   if (type === PaymentType.Monthly) {
-    return new Intl.NumberFormat('en-US', currencyFormatOptions).format(price / 50)
+    return new Intl.NumberFormat('en-US', currencyFormatOptions).format(price / 1000)
```
Lisenish, line 31: Optional: price argument in this function looks misleading, I think it's quantity not a price (because this function actually calculates the price)

src/components/PriceCalculator/index.tsx
```diff
@@ -14,15 +14,15 @@ export default function PriceCalculator() {
     label: numberFormatter.format(entry.value),
     value: entry.value,
   }))
-  selectOptions.push({ label: `${numberFormatter.format(500000)}+`, value: Infinity })
+  selectOptions.push({ label: `${numberFormatter.format(10000000)}+`, value: Infinity })
```
Lisenish, line 17: It was like this before but I don't get why it's manually pushed here in the component, why we can't just add it to the `pricingTable`?
Especially because it seems we do it in the both components where we're using `pricingTable`

Lisenish (commented):
@natapg Good job, couple of minor comments and a question, let's discuss and fix asap but please don't merge it without my singal

content/index/payment-fraud.md
```diff
@@ -82,9 +82,9 @@ blocks:
     ctaText: Try it out
     ctaUrl: https://dashboard.fingerprintjs.com/signup
 inlineCta:
-  title: Create your free account today
-  markdown__Subtitle: "Install our Javascript snippet on your website and start
-    collecting visitorIDs free for up to 1,000 monthly unique visitors. "
+  title: Create your free trial today
```
tan-90, line 85: I think "Start" sounds better than "Create" for the free trial.

tan-90 (approved):

src/pages/pricing/index.tsx
```diff
@@ -79,44 +79,32 @@ function PricingModelsSection({ mainBackground }: { mainBackground: SectionProps
...
           title='On-Demand'
-          description='For on-demand pricing, you pay monthly per unique visitors identified. This plan is recommended for websites with unpredictable or low traffic volumes.'
+          description='For on-demand pricing, you pay monthly per API call. This plan is recommended for users with unpredictable or low traffic volumes.'
           sectionClasses={styles.onDemand}
         >
           <ul className={styles.arrowList}>
-            <li>
-              On-demand is a metered model, where we count your unique visitors and bill you accordingly every month.
-            </li>
-            <li>
```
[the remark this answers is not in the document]
natapg: I shared it in our thread but didn't receive any feedback from Savannah

* The minimum plan is $100/month, which includes 100,000 identifications per month. Unused API calls are not carried over to the next month.

this was the previous message, I don't know if the "minimum plan" is confusing since there is also the free tier, I will ask her

src/helpers/pricing.ts
natapg: you are right, it's not a price, it's the number of API calls / UV

src/components/PriceCalculator/index.tsx
natapg: I think this is because we are using the `pricingTable` value to create the label, so If we add a { label: '10M+', value: Infinity } to the array, it will show an infinity on the selector instead of a "10000000+", so I think it will need some refactor if we want to move it to the array

natapg:
```
const selectOptions = pricingTable.map((entry) => ({
    label: numberFormatter.format(entry.value),
    value: entry.value,
  }))
```

natapg: so an option could be to add this condition to `selectOptions`
```
  const selectOptions = pricingTable.map((entry) => ({
    label: entry.value === Infinity ? `${numberFormatter.format(10000000)}+` : numberFormatter.format(entry.value),
    value: entry.value,
  }))
```

and with that, we can add it to the `pricingTable` and remove both pushes @Lisenish

natapg:
all changes have been addressed @Lisenish

Lisenish (approved):

PR #219, opened 2021-07-09
Add an optional "related blogs" content block to use case pages

natapg:
Closes WEB-160


tan-90 (approved):

static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: feat/WEB-160-add-optional-related-blogs-content-to-use-case
```
Lisenish, line 3: Don't forget to change back

src/templates/static-page-content.module.scss
```diff
@@ -1,5 +1,5 @@
 .section {
-  padding: 2rem 0;
+  padding: 2rem 0 (rem(75px) - 2rem);
```
Lisenish, line 2: Why do we need this arithmetics here and in a file above?

Lisenish (commented):
Good job Nata

src/templates/static-page-content.module.scss
natapg: because in this case we need a 75px padding-bottom according to the mockup, but we already have a 2rem padding from the blog posts so I did the subtraction. I can also put rem(59px) which is the result or what would be the correct way to proceed?

natapg: image
image

PR #223, opened 2021-07-17
account sharing - hero section

natapg:
First Section
Mockup is here: figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI?node-id=0%3A1


src/components/common/Button/index.tsx
```diff
@@ -5,6 +5,7 @@ import classNames from 'classnames'
 interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   variant?: 'primary' | 'outline' | 'clear' | 'faded'
   small?: boolean
+  big?: boolean
```
tan-90, line 8: It's better to have this as `size?: 'small' | 'big'`

tan-90 (approved):
Approving in advance but with one small change request.

src/components/common/Button/index.tsx
natapg: you are right, thanks!

PR #224, opened 2021-07-19
account sharing - how much are you losing section

natapg:
Second Section
Mockup is here: figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI?node-id=0%3A1


src/components/accountSharing/HowMuchLosingSection/HowMuchLosingSection.module.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  @include media('<=phoneLandscape') {
+    padding: 0 0 rem(42px) 0;
+  }
+}
+
+.container {
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  --contentWidth: 1232px;
```
tan-90, line 14: Why do we need this specific override? (hard to see without a DP)

tan-90 (approved):

src/components/accountSharing/HowMuchLosingSection/HowMuchLosingSection.module.scss
natapg: you are right, we don't need it, I can use the 'large' size

PR #226, opened 2021-07-20
Fix some broken links 

natapg:
image


src/components/Breadcrumbs/BreadcrumbsSEO.tsx
```diff
@@ -21,7 +21,7 @@ function getStructuredData(breadcrumbs: Array<Breadcrumb>): string {
       '@type': 'ListItem',
       position: index + 1,
       name: kebabToStart(label),
-      item: `${BASE_URL}${withTrailingSlash(path)}`,
+      item: `${withoutTrailingSlash(BASE_URL)}${withTrailingSlash(path)}`,
```
tan-90, line 24: I did a search and it doesn't look like we use `BASE_URL` that often so this seems a bit overkill. Can't we remove the trailing slash from `BASE_URL` and then only use the `withTrailingSlash` helper?
Or would that require even more changes?

tan-90 (commented):
Are there any situations where we

src/components/Breadcrumbs/BreadcrumbsSEO.tsx
natapg: yea, that was my first option but I didn't want to break anything else. But you are right, it would only be a matter of using withTrailingSlash when necessary, I will make the change

natapg: done,please check @tan-90

tan-90 (approved):

PR #241, opened 2021-07-28
Remove annual discount mentions

natapg:
Closes WEB-191

@tan-90 a little more context so you can also review this PR. These changes were discussed with Savannah internally, they may not look good aesthetically, but the pricing page will be changed soon so she's fine with that


src/helpers/pricing.ts
```diff
@@ -28,12 +26,5 @@ export function calculatePrice(identifications: number, type: PaymentType): stri
...
 
-  if (type === PaymentType.Monthly) {
-    return new Intl.NumberFormat('en-US', currencyFormatOptions).format(identifications / 1000)
-  }
-  if (type === PaymentType.Annually) {
-    return new Intl.NumberFormat('en-US', currencyFormatOptions).format((identifications / 1000) * 0.8)
-  } else {
-    throw new Error('Payment type is required')
-  }
+  return new Intl.NumberFormat('en-US', currencyFormatOptions).format(identifications / 1000)
```
Lisenish, line 29: Minor: please extract this formatter instantiation `new Intl.NumberFormat('en-US', currencyFormatOptions)` to a reusable const variable outside of the funciton, there is no need to create same formatter over and over again on each function call

Lisenish:
We mention 'several' here but in fact now there is only one advantage
image

Lisenish (approved):
Only small things, approved in advance

natapg:
> We mention 'several' here but in fact now there is only one advantage
> image

I agree, and I think it's not worth it to say "has an advantage", I am going to remove the description for this model as I suggested to Savannah

PR #243, opened 2021-07-30
Pricing page redesign

natapg:
Closes WEB-161


src/components/Billing/index.tsx
```diff
@@ -33,13 +33,19 @@ export default function Billing() {
...
   const recalculatePricing = (value: number) => {
-    if (value === Infinity) {
-      setMonthlyPaymentLabel('Custom pricing')
-      return
+    switch (value) {
+      case Infinity:
+        setMonthlyPaymentLabel('Custom pricing')
+        break
+      case freeApiCalls:
+        setMonthlyPaymentLabel('0$')
```
tan-90, line 41: Is it intentional that we have <value>$ for the free tier and then it changes to $<value>?

src/components/PriceCalculator/index.tsx
```diff
@@ -1,140 +1,265 @@
...
-    if (newCustomCount !== '') {
-      const strippedValue = newCustomCount.replace(/\.|,/, '')
-
-      setCustomCount(parseInt(strippedValue, 10))
-    } else {
-      setCustomCount(undefined)
+  const recalculatePricing = (value: number) => {
+    switch (value) {
+      case Infinity:
+        setMonthlyPaymentLabel('Custom Us')
```
tan-90, line 38: Contact Us

src/components/PriceCalculator/index.tsx
```diff
@@ -1,140 +1,265 @@
...
+        break
+      case freeApiCalls:
+        setMonthlyPaymentLabel('0$')
```
tan-90, line 41: Same question as before.

src/components/PriceCalculator/index.tsx
```diff
@@ -1,140 +1,265 @@
...
-      return '$0'
-    } else if (isCustomPricing) {
-      return 'Custom'
-    } else {
-      return getPrice()
-    }
-  }
+  useEffect(() => {
+    recalculatePricing(sliderTable[sliderValue].value)
+  }, [sliderTable, sliderValue])
```
tan-90, line 53: Since `recalculatePricing` is already called on the slider change handler and the slider table won't change isn't the effect unnecessary?

src/components/common/RangeSlider/index.tsx
```diff
@@ -42,22 +45,26 @@ export default function RangeSlider({
...
+        <>
+          <span className={styles.output}>{values[currentValue].label}</span>
+          <label htmlFor='billingSlider' className={styles.label}>
+            {values.map(({ label }) => {
+              return (
+                <span key={`slider_label_${label}`} className={styles.text}>
+                  {label}
+                </span>
+              )
+            })}
```
tan-90, line 61: Minor nitpick: We could return the span directly without a function body to save a return statement.

src/components/common/RangeSlider/index.tsx
```diff
@@ -67,6 +74,22 @@ export default function RangeSlider({
...
+              <span
+                key={`slider_label_${label}`}
+                className={classNames(styles.onlyCurrentText, {
+                  [styles.showValue]: value === values[currentValue].value,
+                })}
+              >
+                {label}
+              </span>
+            )
+          })}
```
tan-90, line 90: Same as last one.

tan-90 (commented):
Only minor things and one possible change, everything else looks good to me.

I know the copy is a WIP but I marked some copy issues just in case.

src/components/Billing/index.tsx
natapg: I will ask Kate, but yea makes sense. When we had the free tier with unique visitors, it was also this way https://deploy-preview-200--fingerprintjscom.netlify.app/

src/components/PriceCalculator/index.tsx
natapg: it is, I've removed it! thanks

natapg: oh, now I see why I added it, it is so that it updates the first value when loading the page, otherwise it shows 20K instead of 0 $ @tan-90

tan-90 (approved):

PR #252, opened 2021-08-06
Add better loading state to demo widget

natapg:
Closes WEB-180


tan-90 (changes requested):
For the skeletons, I think it's better to do the same as react-material-ui and have a component that can take the width/height as props. That way we won't need as many named classes when building the skeletons.

https://material-ui.com/components/skeleton/#wave-example

natapg:
> For the skeletons, I think it's better to do the same as react-material-ui and have a component that can take the width/height as props. That way we won't need as many named classes when building the skeletons.
> 
> [material-ui.com/components/skeleton/#wave-example](https://material-ui.com/components/skeleton/#wave-example)

Added a Skeleton component. Thanks, now it looks cleaner!

tan-90 (approved):

PR #253, opened 2021-08-06
Added lead_source field to form

natapg:


src/pages/contact-sales/index.tsx
```diff
@@ -59,7 +59,7 @@ function TalkToSalesSection() {
     <Section className={styles.form}>
       <h1 className={styles.header}>Talk to an Expert</h1>
       <h2 className={styles.subHeader}>Fill out the form below and we will reach out shortly.</h2>
-      <ContactSalesForm />
+      <ContactSalesForm leadSource='Contact Us Form' />
```
Lisenish, line 62: I understand why you did this but I think you can just extract it to a constant and inline in `ContactSalesForm`.
It's unlikely that we will have some other place where we would need to provide different source

Lisenish (approved):
One small comment, approved in advance

PR #259, opened 2021-08-17
Add customizable CTA

natapg:
Closes WEB-179 & WEB-155


src/components/Content/Content.module.scss
```diff
@@ -167,14 +167,14 @@
 
   summary::after {
     margin-left: rem(14px);
-    content: url(/img/uploads/plus-simbol.svg);
+    content: url('../../img/plus-simbol.svg');
```
natapg, line 170: changed to relative path to fix problem with storybook

static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: feat/WEB-179-add-customizable-CTA-blogs
```
natapg, line 3: I'll change it after Savannah updates CTAs and before merging

tan-90 (approved):
Looks good to me, really nice touch that on mobile you can scroll over the CTA and it will scroll the page below.

static/admin/config.yml
Lisenish: Aren't they optional? Maybe she could update it directly on the prodcution?

src/templates/long-form-content.tsx
```diff
@@ -145,21 +156,34 @@ export function LongFormContentTemplate({
       )}
 
       <Section className={styles.root}>
-        <Container size='small' className={styles.container}>
-          <h1 className={styles.title}>{post.title}</h1>
-          <div className={styles.actionBar}>
-            <ActionBar {...actionBar} />
+        <Container className={styles.container}>
+          <div className={styles.header}>
```
Lisenish, line 160: We're using a lot of div's, let's try to use more semantic tags as well (probably no need to change anything here, but mostly for the future).
@tan-90 FYI for you as well

https://dev.to/kenbellows/stop-using-so-many-divs-an-intro-to-semantic-html-3i9i

src/components/CustomizableCTA/CloseSVG.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="10" height="10" viewBox="0 0 10 10" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Lisenish, line 1: We still didn't have `cross` svg in the websire? Maybe we should make it common and not local then? I think we will probably reuse it if it doesn't look very custom

src/templates/long-form-content.tsx
```diff
@@ -145,21 +156,34 @@ export function LongFormContentTemplate({
...
+          <div className={styles.body}>
+            {heroImage.image && <HeroImageComponent {...heroImage} />}
+            <ContentComponent content={body} className={styles.content} />
+            {customCTA.subHeader && (
+              <div className={styles.ctaMobile}>
+                <CustomizableCTA className={styles.card} {...customCTA} />
+              </div>
+            )}
+          </div>
+          {customCTA.subHeader && (
```
Lisenish, line 182: So we have two instances of CustomizableCTA component? 
I understand that it's for mobile probably but I belive there should be a way to implement it with styles only, since we're using Grid now can't we display everything as we needed with it?

Lisenish (changes requested):

static/admin/config.yml
natapg: they are optional, but we need at least one, because otherwise the graphql query fails, and the one that is set is an example

src/components/CustomizableCTA/CustomizableCTA.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const winMax = document.documentElement.scrollHeight - document.documentElement.offsetHeight
+    const heightToHideFrom = winMax * 0.1
+
+    const winScroll = document.documentElement.scrollTop
+    const changeVisibility = winScroll < heightToHideFrom
+    setIsVisible(changeVisibility)
+  }
+  const onClose = () => setIsClosed(true)
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
natapg: ok 👍🏻

src/components/CustomizableCTA/CloseSVG.svg
natapg: we don't have another, only in the modal that we no longer use, and it is not the same, but yea, it's a good idea to move it to a shared folder

src/components/CustomizableCTA/CustomizableCTA.tsx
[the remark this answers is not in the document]
natapg: done @Lisenish

src/templates/long-form-content.tsx
natapg: done @Lisenish

src/templates/long-form-content.tsx
natapg: done @Lisenish

src/components/CustomizableCTA/CustomizableCTA.tsx
```diff
@@ -0,0 +1,56 @@
...
+    if (hideWhenScroll) {
+      window.addEventListener('scroll', listenToScroll)
+      return () => window.removeEventListener('scroll', listenToScroll)
+    } else return () => setIsVisible(true)
+  })
```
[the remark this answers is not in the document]
tan-90: Sorry about missing it. I probably would have made the same mistake 😓
I'll be more careful.

static/admin/config.yml
Lisenish: Oh yeah makes sense, so you may ask Savannah to add at least one and say that she will be able to add the rest at her own pace later, probably it will be faster for her

src/hooks/useShowOnScroll.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react'
+
+export const useShowOnScroll = () => {
```
Lisenish, line 3: I actually meant something more generic, like this https://usehooks.com/useEventListener/ (but maybe only focused on scroll).
But it's fine also as long as we don't have any other use cases, we may need to refactor it later if we will need to attach some other logic to scrolling

src/components/CustomizableCTA/CustomizableCTA.tsx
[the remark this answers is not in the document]
Lisenish: @tan-90 It's fine, just wanted to make sure that we're doing attentive reviews

Lisenish (approved):

PR #263, opened 2021-08-23
fix: added visitors endpoint

natapg:


src/constants/env.ts
```diff
@@ -31,6 +31,12 @@ export const FPJS_ENDPOINT =
     [NetlifyContext.DeployPreview]: process.env.GATSBY_PREVIEW_FPJS_ENDPOINT,
   }) ?? ''
 
+export const FPJS_VISITORS_ENDPOINT =
+  getContextEnv<string>({
+    [NetlifyContext.Production]: process.env.GATSBY_FPJS_VISITORS_ENDPOINT,
+    [NetlifyContext.DeployPreview]: process.env.GATSBY_PREVIEW_FPJS_VISITORS_ENDPOINT,
+  }) ?? ''
```
Lisenish, line 38: I think we can provide more reasonable default here

Lisenish (approved):

PR #264, opened 2021-08-24
small fixes

natapg:
Closes WEB-202

Changes requested: 

1. Links in Blog post. Copy link is cropped. And when user hover on links they become larger, but should be same size
2. Blog tags' appearance: they should be with 100% round corners. Also I don't like actual hover effect and typography. New design is located in UI kit
3. New Header's buttons for desktop and mobile breakpoints are also in UI kit
4. In Incognito Mode Detection article author avatar's proportions has changed
5. I redesigned related articles hover effect. Also located in Ui kit and in blogpost page mockup
Change the size of the image for the featured article in the blog (see image below, its width is ~450px)


src/components/Post/Post.module.scss
```diff
@@ -83,7 +80,7 @@
 
     .image {
       grid-area: image;
-      height: 100%;
+      width: rem(540px);
```
natapg, line 83: @Lisenish I tested with 450px and it looks different from your image, I think it is 540 the right width (it also now has a small padding of 4px requested by Kate)
 
540
image

450
image

PR #265, opened 2021-08-24
fix: disabled auto-zoom on iOS contact form

natapg:
Closes WEB-207

Tested in browserstack


tan-90 (approved):
Looking through the stack overflow answers, this looks like the best way to fix it without the risk of changing the viewport tag and disabling zoom.

PR #267, opened 2021-08-24
fix: updated homepage banner for soc2

natapg:
Closes WEB-210


tan-90 (approved):

src/components/HeaderBar/HeaderBar.module.scss
```diff
@@ -9,42 +8,50 @@
...
 
-  @include media('<tablet') {
-    font-size: rem(12px);
-  }
+  column-gap: rem(12px);
 
-  @include media('<phone') {
-    font-size: rem(11px);
+  @include media('<tablet') {
+    padding-top: rem(8px);
```
tan-90, line 35: Makes a bigger difference than I thought it would, looks really nice on tablet.

natapg: actually, it's a leftover of the previous design

PR #277, opened 2021-09-02
fix: incorrect term in PSD2 article

makma:
Fix incorrect term in the PSD2 article


natapg (approved):

Lisenish:
@natapg You should've merged it by yourself if you approved it

PR #284, opened 2021-09-07
Fix: bring back lead submit event

natapg:
Closes WEB-220


src/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,26 @@
+import { useState } from 'react'
+
+const useLocalStorage = <T>(key: string, initialValue: T): [string, typeof setValue] => {
```
Lisenish, line 3: Shouldn't be the 1st type in return value tuple be of type of `T`?
```<T>(key: string, initialValue: T): [T, typeof setValue]```

File: src/pages/contact-sales/confirm/index.tsx [hunk 2]
```diff
@@ -22,6 +24,16 @@ export default function ConfirmPage({ pageContext }: ContactSalesPageProps) {
     siteUrl: `${siteMetadata.siteUrl}${pathname}`,
   }
 
+  const [state, setState] = useLocalStorage('sendEvent', true)
```
Lisenish, line 27: Please provide a better names here:
1. `state`/`setState` - should be something like `isLeadSubmitEnabled`/`setLeadSubmitEnabled`
2. The local storage key should be more specific to avoid potential conflicts, let's name it `track_lead_submit_enabled` or something like this.

src/pages/contact-sales/confirm/index.tsx
```diff
@@ -22,6 +24,16 @@ export default function ConfirmPage({ pageContext }: ContactSalesPageProps) {
...
+
+  useEffect(() => {
+    if (state) {
+      setState(false)
+      trackLeadSubmit()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Lisenish, line 34: Please provide explanatory comment why it is disabled and that it's intended

Lisenish (approved):

PR #288, opened 2021-09-08
Demo Page: Algorithm section

natapg:
mock-up: https://figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI-%2B?node-id=0%3A1

DP: https://deploy-preview-286--fingerprintjscom.netlify.app/demo/


src/components/demo/DemoSection/AlgorithmSection/AlgorithmSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+export default function AlgorithmSection({ visits, currentVisit, visitorId }: CurrentVisitProps) {
+  const offset = getTimezoneOffset()
+  const width = window.screen.width
+  const height = window.screen.height
+
+  return (
+    <Section className={styles.root}>
+      <Header />
+      <Container size='large' className={styles.algorithmContainer}>
+        <section className={styles.browserSignalsTitle}>Browser fingerprinting details</section>
```
tan-90, line 31: I think the elements that just hold the title should not be sections, but it's probably best to get @Lisenish input on this.

tan-90 (commented):
This looks like a ton of work, great job!

I'm not quite sure about the use of sections, but @Lisenish is better suited to judge that. Everything else looks good to me.

src/components/demo/DemoSection/AlgorithmSection/AlgorithmSection.tsx
natapg: I agree, I started with just some sections and it ended up getting out of control :P

Lisenish: Yeah Vinicius is right, `h1, h2...` are semantically correct tags for titles

src/components/demo/DemoSection/AlgorithmSection/AlgorithmSection.tsx
```diff
@@ -0,0 +1,205 @@
...
+            icon={<PlanetSVG />}
+            title={`${currentVisit?.ipLocation?.country?.code}, ${currentVisit?.ipLocation?.city?.name} GTM ${offset}`}
+          />
+          <Card icon={<WindowSVG />} title={`resolution ${width}x${height}`} />
+          <Card icon={<MobileSVG />} title={currentVisit?.browserDetails?.os} />
+        </section>
+        <section className={styles.browserRows}>
+          <div className={styles.browserRowsSVG} />
+        </section>
+        <section className={styles.otherSignalsTitle}>Other identifiers</section>
```
Lisenish, line 44: Same here.
When I asked to create more semantic html I didn't meant to just replace all `div`'s with `section`s it doesn't make any sense and doesn't do anything better, you should always try to pick the best sematically suitable tag among all of them, you also have `article`, `footer`, `header`, `nav`, and many other tags

src/components/demo/DemoSection/AlgorithmSection/AlgorithmSection.tsx
```diff
@@ -59,22 +59,22 @@ export default function AlgorithmSection({ visits, currentVisit, visitorId }: Cu
...
           <div className={styles.visitorRowSVG} />
-        </section>
-        <section className={styles.visitorIdTitle}>Your visitor Id</section>
+        </div>
+        <h2 className={styles.visitorIdTitle}>Your visitor Id</h2>
         <section className={styles.visitorId}>
           <Card variant='visitor' icon={<VisitorSVG />} title={visitorId} />
         </section>
-        <section className={styles.mobileRows}>
+        <aside className={styles.mobileRows}>
```
tan-90, line 72: I think aside is more suited for actual content and since this is just a graphic it could be a div or even a span.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside

tan-90 (approved):
Just a minor comment but approving in advance

PR #299, opened 2021-09-14
moved companies logos to components

natapg:
Closes WEB-215


src/components/ClientsSection/index.tsx
```diff
@@ -30,15 +27,29 @@ export default function ClientsSection() {
...
+          <CompanyLogo icon={<WesterUnionSVG className={styles.logo} />} />
+          <CompanyLogo icon={<YahooSVG className={styles.logo} />} />
         </div>
       </Container>
     </Section>
   )
 }
+
+interface CompanyLogoProps {
+  icon: React.ReactNode
```
tan-90, line 51: Instead of typing this as a `ReactNode` you could use `React.FunctionComponent<React.SVGProps<SVGSVGElement>>`. It's a very ugly type, but then you wouldn't need to repeat  `className={styles.logo}` so I think the final result is cleaner.

Here's an example:
https://github.com/fingerprintjs/fingerprintjs.com/blob/53107abebede093e32d6613aea8d32b3bce627aa/src/components/UseCasesSection/index.tsx#L91

tan-90 (changes requested):

src/components/ClientsSection/index.tsx
natapg: ooh you are right, I was looking for a way to do it and couldn't find it, thank you!

tan-90 (approved):

PR #300, opened 2021-09-15
added console log with careers advertisement

natapg:
Closes WEB-205


Lisenish (approved):

src/components/Layout/index.tsx
```diff
@@ -47,6 +47,13 @@ export function LayoutTemplate({ children, siteMetadata, headerBarTitle, headerB
     }
   }, [shouldEnableAnalytics])
 
+  useEffect(() => {
```
Lisenish, line 50: Let's move this effect to a separate file and name it

PR #302, opened 2021-09-16
Demo Page: Requested changes

natapg:
Changes requested by Kate and Valentin


src/components/demo/DemoSection/VisitsSection/VisitsSection.tsx
```diff
@@ -139,13 +139,32 @@ interface ContentProps {
...
+      catchFraudsters: file(relativePath: { eq: "CatchLayers.png" }) {
+        childImageSharp {
+          fluid {
+            ...GatsbyImageSharpFluid
+          }
+        }
+      }
+    }
+  `)
+
```
tan-90, line 160: Do these need to be fluid? As far as I can tell their size never changes and if so they could be fixed.

tan-90 (approved):
One small comment, approving in advance

src/components/demo/DemoSection/VisitsSection/VisitsSection.tsx
natapg: you are right! changing it

PR #306, opened 2021-09-20
feat: Added interactive chart

natapg:
Closes WEB-228


tan-90 (approved):

Lisenish (approved):
Approved with some comments.

Still not sure when to start the animation on the desktop, maybe we should just auto run it as on mobile. 
I would like to know Kate's opinion but she's on vacation, maybe we should ask Sav for feedback

src/components/demo/DemoSection/VisitsSection/catchFraudstersSection.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { useState, useEffect, useRef } from 'react'
```
Lisenish, line 1: Please add TODO regarding renaming after we migrate to Gatsby 3

src/components/demo/DemoSection/VisitsSection/catchFraudstersSection.tsx
```diff
@@ -12,6 +12,9 @@ import { ReactComponent as IncognitoIconSVG } from './IncognitoIconSVG.svg'
 
 import styles from './VisitsSection.module.scss'
 
+// The file name should be CatchFraudstersSection but due to a bug in gatsby
```
Lisenish, line 15: Please mark it as `TODO` as well, so it's clear that it's something we plan to fix later

PR #311, opened 2021-09-22
Gatsby v3 & Gatsby image plugin migration

natapg:


package.json
```diff
@@ -122,5 +128,8 @@
     "build-storybook": "build-storybook",
     "prettier": "prettier -c .",
     "prettier:fix": "prettier --write --loglevel=error ."
+  },
+  "resolutions": {
+    "gatsby/**/svgo": "2.4.0"
```
natapg, line 133: to avoid warnings in the deploy -> this resolution will not be necessary when the new version of svgo is released
https://github.com/gatsbyjs/gatsby/pull/33266#issuecomment-926794537

gatsby-node.js
```diff
@@ -205,7 +205,7 @@ function configureMiniCssExtractPlugin(config) {
 }
 
 exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
-  if (stage === 'build-javascript') {
+  if (stage === 'develop' || stage === 'build-javascript') {
```
natapg, line 208: to avoid errors in development

Lisenish: Could you please elaborate a little bit more what kind of errors we had here? 
I remember you mentioned it but forgot the details, maybe we need to add a comment here?

src/components/common/PreviewCompatibleImage/PreviewCompatibleImage.tsx
```diff
@@ -29,16 +29,16 @@ const PreviewCompatibleImage = ({ className, imageStyle, imageInfo, titleTag, al
 
   const { alt = altTag ?? '', title = titleTag, childImageSharp, image, extension, publicURL } = imageInfo
 
-  if (!!image && typeof image != 'string' && isFluid(image.childImageSharp)) {
-    return <Img className={className} style={style} fluid={image.childImageSharp.fluid} alt={alt} />
+  if (!!image && typeof image != 'string') {
```
Lisenish, line 32: Have you checked it in CMS, is it still working there?

src/components/common/PreviewCompatibleImage/PreviewCompatibleImage.tsx
```diff
@@ -29,16 +29,16 @@ const PreviewCompatibleImage = ({ className, imageStyle, imageInfo, titleTag, al
...
+    return <GatsbyImage image={image} className={className} style={style} alt={alt} />
   }
 
-  if (childImageSharp && isFluid(childImageSharp)) {
+  if (childImageSharp) {
     return (
-      <Img
+      <GatsbyImage
+        image={childImageSharp['gatsbyImageData']}
```
Lisenish, line 39: Why not just `childImageSharp.gatsbyImageData`? Because of the types?

Lisenish (commented):
@natapg Good job, I only quickly reviewed it now but it looks good. 
I will take a look at the UI tomorrow and if everything is good let's deploy tomorrow

gatsby-node.js
natapg: it was giving me errors that the imports were not ordered, so it was ignoring the miniCssExtractPlugin.options.ignoreOrder in develop, I found the solution here
https://github.com/gatsbyjs/gatsby/discussions/30169#discussioncomment-621285
I will add a comment

src/components/common/PreviewCompatibleImage/PreviewCompatibleImage.tsx
natapg: it works, but it gives me this error message: Property 'gatsbyImageData' does not exist on type 'GatsbyImageProps'.

src/components/common/PreviewCompatibleImage/PreviewCompatibleImage.tsx
natapg: yea,all the images are working in the CMS

.prettierrc
```diff
@@ -7,6 +7,6 @@
   "jsxSingleQuote": true,
   "trailingComma": "es5",
   "bracketSpacing": true,
-  "jsxBracketSameLine": false,
+  "bracketSameLine": false,
```
natapg, line 10: jsxBracketSameLine is deprecated
https://prettier.io/blog/2021/09/09/2.4.0.html

src/components/common/PreviewCompatibleImage/PreviewCompatibleImage.tsx
Lisenish: Probably because you needed to change `ImageInfo` type in the beginning of the file since we changed image plugin?

Currently it's the same as if you have wrote `(childImageSharp as any).gatsbyImageData` so it means we're not using types anymore, it's a bad practice since in case if we don't need types we can just write on JS instead of TS.

Lisenish (approved):
I approved it but please address two last comments first

PR #314, opened 2021-09-23
minor fixes

natapg:


tan-90 (approved):

tan-90:
Approved in advance but I see the email is still being decoded, have we decided how to handle that?

PR #317, opened 2021-09-28
Gatsby v3 migration -> Upgrade missing dependencies

natapg:


src/pages/contact-sales/PageContent.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
```
natapg, line 1: to use [Fast Refresh](https://www.gatsbyjs.com/docs/reference/local-development/fast-refresh) the number of exports per page is limited to 1, so I had to move these components to different files

Lisenish: Correct me if I'm wrong but our current Gatsby version uses React 17, so you can omit React imports

PR #320, opened 2021-10-01
fix: added horizontal scrolling to tablet / mobile view related articles

natapg:
Closes WEB-223


src/styles/global-styles.scss
```diff
@@ -137,3 +137,9 @@ img {
   width: 100%;
   height: auto;
 }
+
+// In Safari and iOS the img element is placed above the wrapper hiding the curved borders
+.gatsby-image-wrapper {
+  position: relative;
+  z-index: 0;
```
natapg, line 144: we have this problem on iOS @tan-90 https://stackoverflow.com/questions/66835241/gatsby-image-plugin-not-displaying-properly-on-ios-devices

tan-90 (approved):

PR #324, opened 2021-10-05
Add blog redirect

natapg:


netlify.toml
```diff
@@ -2,5 +2,11 @@
   from = "http://www.fingerprintjs.com"
   to = "https://fingerprintjs.com/"
 
+[[redirects]]
+  from = "/blog/wallpaper-id/"
+  to = "/blog/how-android-wallpaper-images-threaten-privacy/"
+  status = 301
```
tan-90, line 8: Netlify says the redirect status is 301 by default, so we probably don't need it here.

tan-90 (approved):
Just one minor comment, approving in advance.

PR #336, opened 2021-10-24
Added landing page, visited pages and last page fields

natapg:
Closes [WEB-199](https://fpjs.myjetbrains.com/youtrack/issue/WEB-199)

Added [plugin-catch-links](https://www.gatsbyjs.com/plugins/gatsby-plugin-catch-links/) to prevent page refresh on markdown links

The console.log on the contact sales page is for testing only

The sending of these fields will be implemented in the next PR


src/context/HistoryListener.tsx
```diff
@@ -0,0 +1,44 @@
...
+          historyData.visitedPages.splice(0, 1)
+        }
+      }
+    })
+  }, [])
+
+  return <>{children}</>
+}
+
+export const getTrackingData = () => {
```
Lisenish, line 38: Minor: We usually use function declaration everywhere, let's change it for the consistency. 
Also we can just return `historyData` here, since it has exacly the same shape (in the future if we decide to change it we can always map ot here)

yarn.lock
```diff
@@ -9547,10 +9547,10 @@ gatsby-plugin-breadcrumb@^12.1.1:
     prop-types "^15.7.2"
     wildcard-match "^5.1.0"
 
-gatsby-plugin-catch-links@^4.0.0:
-  version "4.0.0"
-  resolved "https://registry.yarnpkg.com/gatsby-plugin-catch-links/-/gatsby-plugin-catch-links-4.0.0.tgz#8d015e36a752f4f878cad5bc313268b48eeeb327"
-  integrity sha512-KSJRsJpwf8i3BV+mssj06OqP9gqi2dx1h/HU/CW1uC/wkup7qxBFUJHd5qCNsuJMKrr0MuBa9rtGHi96ra4w/Q==
+gatsby-plugin-catch-links@^3.14.0:
+  version "3.14.0"
```
Lisenish: Why it was downgraded? Next major version for Gatsby 4?

Lisenish (approved):

yarn.lock
natapg: This version is not compatible with versions less than gatsby v4 (it threw that error during the build) so I downgraded it to 3.14

src/context/HistoryListener.tsx
natapg: in the other PR I changed it to a custom hook so I think it's ok to use a function expression here (as we use in the others custom hooks)

Lisenish:
@natapg Are you waiting for the other branch here? Can't we merge it without it?

natapg:
> @natapg Are you waiting for the other branch here? Can't we merge it without it?

merging this will stop sending the UTM parameters in the query string

PR #345, opened 2021-10-26
new fields : added sales and signup flows

natapg:


src/context/HistoryListener.tsx
```diff
@@ -32,13 +47,46 @@ export function HistoryListener({ children }: { children: React.ReactNode }) {
...
 
-export const getTrackingData = () => {
-  return {
-    landingPage: historyData.landingPage,
-    visitedPages: historyData.visitedPages,
-    previousPage: historyData.previousPage,
+export const useViewTracking = () => {
+  const { isNonEuUser } = useUserLocation()
+
+  if (isNonEuUser) {
```
natapg, line 72: @Lisenish My idea here is to send the cookie data if it is a non-EU user, but if it is an EU user, we still send the available information to salesforce through the lead form and to the API through the embedded form.

src/context/HistoryListener.tsx
```diff
@@ -40,13 +47,46 @@ export function updateHistory(path: string, storeCookies?: boolean) {
...
+      setCookie(utmParamsCookie, JSON.stringify(historyData.utmParams), 30)
+
+      return globalHistory.listen(({ action }) => {
+        if (action === 'PUSH') {
+          setCookie(visitedPagesCookie, JSON.stringify(historyData.visitedPages), 30)
+        }
+      })
+    } else {
+      return
+    }
```
Lisenish, line 63: Looks like useless `else` block, I think we can just remove it? Or you can move it to the top and check `if (!isEuUser) { return }` and put the big block outside of the if

src/context/HistoryListener.tsx
```diff
@@ -40,13 +47,46 @@ export function updateHistory(path: string, storeCookies?: boolean) {
     })
   }, [])
 
+  useEffect(() => {
+    if (isEuUser === false) {
+      setCookie(visitedPagesCookie, JSON.stringify(historyData.visitedPages), 30)
+      setCookie(landingPageCookie, historyData.landingPage, 30)
+      setCookie(utmParamsCookie, JSON.stringify(historyData.utmParams), 30)
```
Lisenish, line 54: Do we need 3 different cookies? Why not just store the full object here?

src/context/HistoryListener.tsx
```diff
@@ -1,37 +1,44 @@
...
+  const { isEuUser } = useUserLocation()
+
+  useEffect(() => {
+    const urlParamsEntries = new URLSearchParams(globalHistory.location.search).entries()
+    const utmParams = objectFromEntries(urlParamsEntries)
+    historyData.utmParams = getUtmParams(utmParams, { referral_url: document.referrer })
+
+    historyData.landingPage = globalHistory.location.pathname
+    historyData.visitedPages.push(globalHistory.location.pathname)
+    return globalHistory.listen(({ action }) => {
```
Lisenish:
```suggestion
    historyData.visitedPages.push(globalHistory.location.pathname)

    return globalHistory.listen(({ action }) => {
```

src/context/HistoryListener.tsx
```diff
@@ -1,37 +1,44 @@
...
-      historyData.visitedPages.splice(0, 1)
-    }
-  }
-  storeCookies && setCookie('visitedPages', JSON.stringify(historyData.visitedPages), 30)
-}
+export function HistoryListener({ children }: { children: React.ReactNode }) {
+  const { isEuUser } = useUserLocation()
+
+  useEffect(() => {
+    const urlParamsEntries = new URLSearchParams(globalHistory.location.search).entries()
```
Lisenish, line 32: I'm not sure if it's correct that we're using `globalHistory` everywhere now and not `location` from the reach custom hook. Probably it's fine, it works?

src/context/HistoryListener.tsx
```diff
@@ -40,13 +47,46 @@ export function updateHistory(path: string, storeCookies?: boolean) {
...
+    if (isEuUser === false) {
+      setCookie(visitedPagesCookie, JSON.stringify(historyData.visitedPages), 30)
+      setCookie(landingPageCookie, historyData.landingPage, 30)
+      setCookie(utmParamsCookie, JSON.stringify(historyData.utmParams), 30)
+
+      return globalHistory.listen(({ action }) => {
+        if (action === 'PUSH') {
+          setCookie(visitedPagesCookie, JSON.stringify(historyData.visitedPages), 30)
+        }
+      })
```
Lisenish, line 60: Do we need two listeners for it? You can set it in the previous listener, no? Or you didn't want to check for region there?

src/context/HistoryListener.tsx
```diff
@@ -1,37 +1,44 @@
...
+import React, { useEffect } from 'react'
+import { globalHistory } from '@reach/router'
+import { getUtmParams } from '../hooks/useUtmParams'
+import { objectFromEntries } from '../helpers/common'
+import { setCookie, getCookie } from '../helpers/cookies'
+import { useUserLocation } from '../hooks/useUserLocation'
+
+const visitedPagesCookie = 'visitedPages'
+const landingPageCookie = 'landingPage'
+const utmParamsCookie = 'utmParams'
```
Lisenish, line 10: We should have only one cookie, also let's name it in some less obvious more and common way so it's not very suspicious for users, e.g. `fpjsWebsiteData`. We can later use it for something else as well

src/context/HistoryListener.tsx
```diff
@@ -40,13 +47,46 @@ export function updateHistory(path: string, storeCookies?: boolean) {
...
-export const getTrackingData = () => {
+export const useViewTracking = () => {
+  const { isEuUser } = useUserLocation()
+
+  if (isEuUser === false) {
+    // We use cookies if the user is not from the EU, but if it is, we still want to send the available data to SF and the MGMT API
+    const landingPage = getCookie(landingPageCookie)
+    const visitedPages = getCookie(visitedPagesCookie)
+    const visitedPagesArray: string[] = visitedPages ? JSON.parse(visitedPages) : historyData.visitedPages
+    const utmParams = getCookie(utmParamsCookie)
```
Lisenish, line 77: You're reading and parsing cookie multiple times here: 
1. It's expensive to do it on each render
2. It's a side effect and you shouldn't do that in render function. Take a look at this example https://medium.com/swlh/react-hooks-usecookie-hook-26ac06ff36b0. Or https://github.com/reactivestack/cookies/tree/master/packages/react-cookie (we don't need necessary add this as a dep, just an example)

Lisenish (changes requested):

src/context/HistoryListener.tsx
```diff
@@ -40,13 +47,46 @@ export function updateHistory(path: string, storeCookies?: boolean) {
...
+  }, [isEuUser])
+
   return <>{children}</>
 }
 
-export const getTrackingData = () => {
+export const useViewTracking = () => {
+  const { isEuUser } = useUserLocation()
+
+  if (isEuUser === false) {
```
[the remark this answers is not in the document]
natapg: I tried that, but during rendering the value is undefined so the condition doesn't pass and it tries to get the cookies giving the following error
`error "document" is not available during server side rendering`

src/context/HistoryListener.tsx
natapg: yeah, it works! tested several times using utm params

src/context/HistoryListener.tsx
natapg: I didn't want to check the region there and add it to the deps,I want to prevent the effect from running and add duplicates to the array

src/context/HistoryListener.tsx
natapg: I've added a useEffect in this case, so it should just run once

src/components/GetStartedForm/index.tsx
```diff
@@ -41,7 +41,6 @@ export default function GetStartedForm({ className }: GetStartedFormProps) {
         fpjsVisitorId: visitorId,
         utmInfo: utmParams,
         landingPage: landingPage,
-        lastPage: visitedPages[visitedPages.length - 1],
```
natapg: @Lisenish I'm going to add this value to the table from the API taking the last element of the visitedPages array in this way I don't have to send it here or from the dashboard

src/context/HistoryListener.tsx
```diff
@@ -1,22 +1,35 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 import { globalHistory } from '@reach/router'
+import { getUtmParams } from '../hooks/useUtmParams'
+import { objectFromEntries } from '../helpers/common'
+import { setCookie, getCookie } from '../helpers/cookies'
+import { useUserLocation } from '../hooks/useUserLocation'
+
+const fpjsWebsiteData = 'fpjsWebsiteData'
```
Lisenish, line 8:
```suggestion
const trackingDataCookieName = 'fpjsWebsiteData'
```

src/context/HistoryListener.tsx
```diff
@@ -32,13 +45,41 @@ export function HistoryListener({ children }: { children: React.ReactNode }) {
     })
   }, [])
 
+  useEffect(() => {
+    if (isEuUser === false) {
```
Lisenish, line 49: I already wrote about nesting, it's better to invert this condition, probably I wans't clear enough I will show to you on our call

src/context/HistoryListener.tsx
```diff
@@ -1,22 +1,35 @@
...
   previousPage: '',
+  utmParams: {},
 }
 
 const visitedPagesLimit = 20
 
 export function HistoryListener({ children }: { children: React.ReactNode }) {
+  const { isEuUser } = useUserLocation()
+
   useEffect(() => {
```
Lisenish, line 29: Please add comments for each effect, to clarify what they're responsible for

src/context/HistoryListener.tsx
```diff
@@ -32,13 +45,41 @@ export function HistoryListener({ children }: { children: React.ReactNode }) {
...
+  const [history, setHistory] = useState(historyData)
+
+  useEffect(() => {
+    if (isEuUser === false) {
+      // We use cookies if the user is not from the EU, but if it is, we still want to send the available data to SF and the MGMT AP
+      const historyDataCookies = getCookie(fpjsWebsiteData)
+      const history: HistoryProps = historyDataCookies ? JSON.parse(historyDataCookies).historyData : historyData
+      setHistory(history)
+    }
+    setHistory(historyData)
```
Lisenish, line 82: We are always overwriting history with historyData here?

src/context/HistoryListener.tsx
```diff
@@ -32,13 +45,41 @@ export function HistoryListener({ children }: { children: React.ReactNode }) {
...
+export const useViewTracking = () => {
+  const { isEuUser } = useUserLocation()
+
+  const [history, setHistory] = useState(historyData)
+
+  useEffect(() => {
+    if (isEuUser === false) {
+      // We use cookies if the user is not from the EU, but if it is, we still want to send the available data to SF and the MGMT AP
+      const historyDataCookies = getCookie(fpjsWebsiteData)
+      const history: HistoryProps = historyDataCookies ? JSON.parse(historyDataCookies).historyData : historyData
```
Lisenish, line 79: I think we discussed that we doesn't need to read it from the cookies on the website side?

Lisenish (changes requested):
@natapg Let's discuss src/context/HistoryListener.tsx on our call tomorrow

src/context/HistoryListener.tsx
```diff
@@ -5,7 +5,7 @@ import { objectFromEntries } from '../helpers/common'
 import { setCookie, getCookie } from '../helpers/cookies'
 import { useUserLocation } from '../hooks/useUserLocation'
 
-const fpjsWebsiteData = 'fpjsWebsiteData'
+const trackingDataCookieName = 'fpjsWebsiteData'
```
Lisenish, line 8: There is an error above

Lisenish (approved):

PR #361, opened 2021-11-08
fix: show validation error in contact sales form

natapg:
Closes WEB-251


src/components/ContactSalesForm/index.tsx
```diff
@@ -23,15 +23,24 @@ export default function ContactSalesForm() {
...
     updateFormState(FormState.Loading)
 
-    function onError() {
+    function onError(error?) {
       updateFormState(FormState.Failed)
+
+      if (error?.message) {
+        if (error?.data[0]?.keyword === 'format') {
+          updateErrorMessage(`Validation error for ${error.data[0].params?.format}`)
+        } else {
```
tan-90, line 39: The `format` property is the name of the rule we used for validation. To get the name of the field we should use `instancePath`. It separates sub properties and array elements using `/`, but since on this form the only value with sub properties is the utm info, we can probably just remove the starting / and use what's left.

(For example if there's a validation error on the email, `instancePath` will be `/email`, but if it was an array of emails and the second email had an error, it would be `/email/1` or `/user/email` for objects)

https://ajv.js.org/api.html#error-objects

tan-90 (changes requested):

src/components/ContactSalesForm/index.tsx
tan-90: We should also check if the error object will have a message, since that will usually be more descriptive:

```
error: {
              data: [
                {
                  instancePath: '',
                  keyword: 'required',
                  message: "must have required property 'emails'",
                  params: {
                    missingProperty: 'emails',
                  },
                  schemaPath: '#/required',
                },
              ],
              message: 'Validation error',
            },
```

natapg: for example if I send an email like emai@localhost the error message is -> "must match format \"email\""
should I show this message insted of Validation error for email? what do you think?

natapg: it could be something like this 
```
      if (error?.message) {
        if (error?.data[0]?.message) {
          updateErrorMessage(error?.data[0]?.message)
        } else if (error?.data[0]?.instancePath) {
          updateErrorMessage(`Validation error for ${error?.data[0]?.instancePath.substring(1)}`)
        } else {
          updateErrorMessage(error?.message)
        }
      }
```

tan-90: Good point, in this case the message from AJV is not that great. I think we can just change it to get the field name from the instance path and leave the message as is.

tan-90 (approved):

PR #363, opened 2021-11-09
fix: include subdomains in setCookie

natapg:


src/helpers/cookies.ts
```diff
@@ -1,11 +1,14 @@
...
 export function setCookie(name: string, value: string, days: number) {
+  const domain = BASE_URL.replace(/^http(s?):\/\//i, '')
   let expires = ''
   if (days) {
     const date = new Date()
     date.setDate(date.getDate() + days)
     expires = '; expires=' + date.toUTCString()
   }
-  document.cookie = name + '=' + (value || '') + expires + '; path=/'
+  document.cookie = name + '=' + (value || '') + expires + '; domain=.' + domain + ' ; path=/'
```
Lisenish, line 11: 1. You don't need leading dot for the domain attribure, it's obsolette now (see https://stackoverflow.com/a/23086139)
1. Please use `??` instead of `||`

Lisenish (approved):

src/helpers/cookies.ts
natapg: 1. I read that too, and I tried without the dot and didn't work for me, but it was locally, I will try in the DP

PR #367, opened 2021-11-11
[Solutions Page] Solutions Overview Page

natapg:
DP: https://deploy-preview-366--fingerprintjscom.netlify.app/solutions/
Mock-up: https://www.figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI-%2B?node-id=1976%3A13990
Ticket: https://fpjs.myjetbrains.com/youtrack/issue/WEB-255

*Using blogs for testing, the CMS integration will be rolled out in the next PR*


src/components/PostGrid/PostGrid.tsx
```diff
@@ -27,73 +30,154 @@ export default function PostGrid({
...
+
+    setSelectedTags(newSelectedTags)
+    const filteredSolutions = posts.filter((solution) => solution.tags?.some((tag) => newSelectedTags.has(tag)))
+    newSelectedTags.size === 0 ? setNumberOfSolutions(posts.length) : setNumberOfSolutions(filteredSolutions.length)
 
-      <div className={styles.posts}>
-        {name && (
+    setTimeout(() => {
+      newSelectedTags.size === 0 ? setSelectedSolutions(posts) : setSelectedSolutions(filteredSolutions)
+    }, timeout)
```
borislobanov92, line 65: you can avoid some duplication in the last section:
```
const filteredSolutions = newSelectedTags.size === 0 ? posts : posts.filter((solution) => solution.tags?.some((tag) => newSelectedTags.has(tag)))

setNumberOfSolutions(filteredSolutions.length)
setTimeout(() => {
  setSelectedSolutions(filteredSolutions)
}, timeout)
```

src/components/PostGrid/PostGrid.tsx
```diff
@@ -27,73 +30,154 @@ export default function PostGrid({
...
-  ) : (
-    <div className={styles.root}>
-      {tags && (
-        <div className={styles.tags}>
-          <h2>Tags</h2>
-          <TagList tags={tags} direction='vertical' format='title' />
-        </div>
-      )}
+  const [selectedTags, setSelectedTags] = useState(new Set())
+  const [selectedSolutions, setSelectedSolutions] = useState(posts)
```
borislobanov92, line 34: this is a matter of taste but I think 'selected' is ok for tags because the user manually selects them but for solutions I'd use 'filtered' (as you did below in the  method). I believe it better reflects what that variable actually is

src/components/PostGrid/PostGrid.tsx
```diff
@@ -1,13 +1,15 @@
...
 import classNames from 'classnames'
 import TagList from '../TagList/TagList'
 import { Swiper, SwiperSlide } from 'swiper/react'
 import { Pagination } from 'swiper'
+import { kebabToTitle } from '../../helpers/case'
 
 import styles from './PostGrid.module.scss'
 
 export interface PostGridProps {
+  variant?: 'posts' | 'solutions'
```
borislobanov92, line 12: I'd use enum for this instead of 'magic' strings.

src/components/PostGrid/PostGrid.tsx
```diff
@@ -27,73 +30,154 @@ export default function PostGrid({
...
+              </div>
+            )}
+            <div className={classNames(styles.grid, { [styles.threePerRow]: perRow === 'three' })}>
+              {posts.map((post) => {
+                return <Post perRow={perRow} key={post.path} limitTextLines={limitPostLines} {...post} />
+              })}
+            </div>
+          </div>
+        </div>
+      )
```
borislobanov92, line 138: this is too much code and it's hard to read, please refactor these into separate  local components (like Posts and SwiperPosts or smth), this  way the ternary will look much better

src/components/PostGrid/PostGrid.tsx
```diff
@@ -27,73 +30,154 @@ export default function PostGrid({
...
+                {tags?.map((tag) => (
+                  <li key={tag} className={styles.item} onClick={() => handleSelectedTags(tag)}>
+                    <span className={classNames(styles.tag, { [styles.selectedTag]: selectedTags.has(tag) })}>
+                      {kebabToTitle(tag)}
+                    </span>
+                  </li>
+                ))}
+              </ul>
+            </section>
+          )}
```
borislobanov92, line 178: actually, I don't see the reason to have the Solutions inside if of the PostGrid. It's great  that  you decided to  reuse thing but PostGrid is the building block and the Solutions is the business logic component. So ideally you should have two business components: Posts and Solutions and both of the inside should use a components named something like ContentGrid (a more generic name). otherwise we are putting logic into a presentational component, which is harder to maintain and reuse. with this approach if we  are working in a team and I need to build a grid based page I'll have to add another condition and rendering logic to this component, it will end up being a huge mess. what I'd want to  do instead is to take the Grid component, pass the data to  it  and  it will render the data  according to the settings.  let  me know if something is unclear, we  can get on a call and discuss.

src/components/PostGrid/PostGrid.tsx
```diff
@@ -27,73 +30,154 @@ export default function PostGrid({
...
+              <span className={styles.showingSolutions}>
+                Showing {numberOfSolutions} solutions
+                {selectedTags.size > 0 && ` matching ${selectedTags.size} filter`}
+              </span>
+              <span
+                className={classNames(styles.filter, { [styles.hideFilter]: selectedTags.size === 0 })}
+                onClick={() => handleSelectedTags()}
+              >
+                Clear all filters
+              </span>
```
borislobanov92, line 153: please use <button> element for buttons instead of spans

src/components/PostGrid/PostGrid.tsx
```diff
@@ -27,73 +30,154 @@ export default function PostGrid({
...
+        </div>
+      )
+    case 'solutions': {
+      return (
+        <div className={styles.solutionsRoot}>
+          <section className={styles.posts}>
+            <div className={styles.filterSection}>
+              <span className={styles.showingSolutions}>
+                Showing {numberOfSolutions} solutions
+                {selectedTags.size > 0 && ` matching ${selectedTags.size} filter`}
```
borislobanov92, line 146: 'filter' is always singular, '18 solutions matching 3 filter' looks awkward. also, I think we should have some native  speaker proofread this line, I'm  not  sure if 3 tags are considered 3 filters, to me it's one filtering  systems, so maybe we should change the wording here

src/components/solutions/HeroSection/HeroSection.tsx
```diff
@@ -0,0 +1,25 @@
...
+          <p className={styles.caption}>Solutions</p>
+          <h1 className={styles.title}>Build your own anti-fraud stack</h1>
+          <p className={styles.description}>
+            Virtually all online fraud can be prevented with accurate visitor identification. FingerprintJS makes
+            designing effective anti-fraud workflows easy for any developer team.
+          </p>
+        </section>
+        <section className={styles.imageSection}>
+          <HeroSvg className={styles.image} />
+        </section>
```
borislobanov92, line 21: not sure if sections are correctly used here, usually the image and the description are one unit and you already have a section wrapper on top, so I'd probably just use divs here

borislobanov92 (changes requested):

src/components/PostGrid/PostGrid.tsx
natapg: yeah, I agree, my idea is to move all the new logic into a new component SolutionGrid (because in this component we have the filter logic that we don't need in the post grid) but still reuse the Post component there because we have the same design

src/components/PostGrid/PostGrid.tsx
natapg: good point, I didn't notice the singular form there, I will talk to Kate and Sav about this

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import classNames from 'classnames'
+
+import styles from './Grid.module.scss'
+
+export interface GridProps {
+  items: Array<React.ReactNode>
+  perRow?: 'four' | 'three'
```
borislobanov92, line 8: I think it would be more logical if this was just a number type and limit it (but also have a default value)
```
type PerRow = 3 | 4
```

src/components/Posts/Posts.tsx
```diff
@@ -0,0 +1,144 @@
...
+import Grid from '../Grid/Grid'
+
+import styles from './Posts.module.scss'
+
+export interface PostsProps {
+  posts: Array<PostProps>
+  name?: string
+  link?: React.ReactNode
+  tags?: string[]
+  perRow?: 'four' | 'three'
```
borislobanov92: I suggest that you create a separate type for it next to the Grid component and import it from there, so this interface looks  like
```
export type PostsProps = PerRow & {
  posts: Array<PostProps>
  name?: string
  link?: React.ReactNode
  tags?: string[]
  ...
}
```

src/components/Posts/Posts.tsx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+interface PostGridProps {
+  posts: Array<PostProps>
+  name?: string
+  link?: React.ReactNode
+  tags?: string[]
+  perRow?: 'four' | 'three'
+  nameIsCentered?: boolean
+  limitPostLines?: boolean
```
borislobanov92, line 117: there are several similarities between  this interface and SwiperPostsProps, please create a shared type and extend it in  both places

src/components/solutions/Solution/Solution.tsx
```diff
@@ -0,0 +1,132 @@
...
+  if (editing) {
+    const solution: SolutionProps = {
+      title: '',
+      description: '',
+      publishDate: dateFormatter.format(Date.now()),
+      path: '/',
+      tags: [],
+    }
+
+    if (data.frontmatter) {
```
borislobanov92, line 72: we already have this check at the  top of the method

src/components/solutions/SolutionsSection/SolutionsSection.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <Container size={'large'} className={styles.gridContainer}>
+      <div className={styles.solutionsRoot}>
+        <section>
+          <div className={styles.filterSection}>
+            <span className={styles.showingSolutions}>
+              Showing {numberOfSolutions} {numberOfSolutions === 1 ? 'solution' : 'solutions'}
+              {selectedTags.size > 0 &&
+                ` matching ${selectedTags.size} ${selectedTags.size === 1 ? 'filter' : 'filters'}`}
```
borislobanov92, line 61: Optional: it's better to make a helper for this,  like here this one https://github.com/fingerprintjs/dashboard/blob/2fa81f5ed34d4f50c449dc5459a976715b9421ad/src/helpers/pluralize.ts

src/components/solutions/SolutionsSection/SolutionsSection.tsx
```diff
@@ -0,0 +1,99 @@
...
+              Clear all filters
+            </span>
+          </div>
+          <div onTransitionEnd={() => setFade(true)}>
+            <Grid
+              items={filteredSolutions.map((solution) => {
+                return <Solution key={solution.path} {...solution} />
+              })}
+              perRow='three'
+              className={classNames(styles.solutionsGrid, { [styles.fadeOut]: !fade }, { [styles.fadeIn]: fade })}
```
borislobanov92, line 76: you can have  one object
```
{ 
  [styles.fadeOut]: !fade,
  [styles.fadeIn]: fade,
}
```

borislobanov92 (approved):

PR #377, opened 2021-11-15
Upgrade to Gatsby v4

natapg:


Lisenish:
@natapg Wow, no breaking changes for us so we even don't need to change anything in the source code? Nice. 
Have you checked storybook?

Lisenish (approved):

natapg:
> @natapg Wow, no breaking changes for us so we even don't need to change anything in the source code? Nice. Have you checked the storybook?

yeah, the breaking changes in this version do not affect us. Storybook works well.
The only thing is that some community plugins are not updated to accept this new version of gatsby in their peerDependencies but they work fine, so I will open PRs to add it

PR #381, opened 2021-11-18
fix: removed preprocessMarkdown from gatsby node

natapg:


src/templates/case-study-content.tsx
```diff
@@ -210,16 +208,7 @@ function mapToSummary(querySummary: QuerySummary, preview = false): SummaryProps
...
-              <MarkdownContent
-                markdown={
-                  result?.markdown__Content ??
-                  'Sed ut fermentum dolor. Vivamus pulvinar nisi leo, in accumsan diam pretium id. Vestibulum aliquam posuere enim, sed finibus sapien fringilla pharetra. Ut sollicitudin nunc non dui placerat facilisis. Duis neque turpis, dictum sit amet sagittis ut, finibus ac eros. Cras pulvinar laoreet diam vel lacinia.'
-                }
-              />
-            ) : (
-              <DangerouslyRenderHtmlContent content={result?.markdown__Content ?? ''} />
-            ),
+            children: <DangerouslyRenderHtmlContent content={result?.markdown__Content ?? ''} />,
```
Lisenish, line 211: Does it work? DangerouslyRenderHtmlContent expects html content and not markdown right?

Lisenish (approved):
I agree to deploy it but only as a temporary hot fix workaround. 
I clearly saw the empty space in the places with the markdown text and only after about a second I was able to see the rendered text

src/templates/case-study-content.tsx
natapg: yeah, here we are sending the html content

PR #382, opened 2021-11-18
fix: added markdown custom Graphql schema

natapg:
based on this https://www.gatsbyjs.com/docs/reference/graphql-data-layer/schema-customization/ I'm defining a md tag to mark markdown fields


tan-90 (approved):
This is very clean, great job!

I think we could even remove the `markdown__` prefix, originally it was only meant to mark fields that should be converted.

On vscode you could just do a global find replace with `markdown__([A-Z])` -> `\l$1`

natapg:
> This is very clean, great job!
> 
> I think we could even remove the `markdown__` prefix, originally it was only meant to mark fields that should be converted.
> 
> On vscode you could just do a global find replace with `markdown__([A-Z])` -> `\l$1`

nice ! I think I would have wasted a lot of time doing it differently, thanks for the advice

PR #388, opened 2021-11-26
Feat: authors page

natapg:
Closes [WEB-175](https://fpjs.myjetbrains.com/youtrack/issue/WEB-175)


src/templates/author.tsx
```diff
@@ -0,0 +1,118 @@
...
+            <Breadcrumbs breadcrumbs={breadcrumbs.slice(1)} />
+          </Container>
+        </>
+      )}
+
+      <Section className={styles.authorSection}>
+        <Container size='large'>
+          <h1 className={styles.aboutAuthor}>About {author} </h1>
+          <div className={styles.authorSummary}>
+            <section>
```
tan-90, line 54: Same as the previous PR, if possible it would be nice to use the `Section` component here and keep the `Container` outside.

src/templates/author.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  return (
+    <LayoutTemplate siteMetadata={siteMetadata}>
+      {breadcrumbs && (
+        <>
+          <BreadcrumbsSEO breadcrumbs={breadcrumbs} />
+          <Container size='large'>
+            <Breadcrumbs breadcrumbs={breadcrumbs.slice(1)} />
+          </Container>
+        </>
```
tan-90, line 47: IIRC it doesn't matter where the `BreadcrumbsSEO` component is rendered so we could put it inside the `Container` here to save a fragment.

tan-90 (approved):

PR #402, opened 2021-12-02
feat: added demo section to hero

natapg:
Closes [WEB-279](https://fpjs.myjetbrains.com/youtrack/issue/WEB-279)


src/hooks/useBotDResponse.ts
```diff
@@ -0,0 +1,29 @@
...
+    async function getVisitorData() {
+      try {
+        const botD = await Botd.load({
+          token: clientToken,
+          mode: 'allData',
+        })
+        const result = await botD.detect()
+        setVisitorData(result)
+      } catch (error) {
+        // noop
```
natapg, line 21: not sure what to do in case of an error, just log the error message or use rollbar maybe?

src/styles/vendors/tippy.scss
```diff
@@ -69,7 +69,6 @@
 .tippy-content {
   position: relative;
   padding: 1rem 2rem;
-  z-index: 1;
```
natapg, line 72: removing this property because is interfering with the tippy arrow

src/hooks/useBotDResponse.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState } from 'react'
+import Botd from '@fpjs-incubator/botd-agent'
+import { BOTD_TOKEN } from '../constants/env'
+
+export const useBotDResponse = () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Lisenish, line 6: Let's leave a  TODO here to add types (I believe it's temporary until they didn't export types, correct?)

src/hooks/useBotDResponse.ts
```diff
@@ -0,0 +1,29 @@
...
+  const [visitorData, setVisitorData] = useState<any>()
+
+  const clientToken = BOTD_TOKEN
```
Lisenish, line 9: That's the purpose? You can just inline `BOTD_TOKEN`

src/hooks/useBotDResponse.ts
Lisenish: It's better to return it from the hook (see how react-query does it for example). You can return `isError` boolean property and error object.
What should we do depends on the errors, what are possible errors? Does BotD team have any guidelines regarding it, types for errors?

In this branch you can just return it, we can decided what to do/show on UI when you discussed the types of the errors with BotD team

src/styles/vendors/tippy.scss
Lisenish: Thanks for the comment. Are you sure it wouldn't break anything? (have you checked?) I wonder why it was here in the first place then

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -1,13 +1,83 @@
...
+  searchEngine: boolean
+  vm: boolean
+}
+
 export default function HeroSection() {
+  const visitorData = useBotDResponse()
+
+  const [isLoading, setIsLoading] = useState(true)
+  const [isBot, setIsBot] = useState(false)
+  const [detectedBots, setdetectedBots] = useState<DetectedBots>({
```
Lisenish, line 37:
```suggestion
  const [detectedBots, setDetectedBots] = useState<DetectedBots>({
```

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -1,13 +1,83 @@
...
+  automationTool: boolean
+  browserSpoofing: boolean
+  searchEngine: boolean
+  vm: boolean
+}
+
 export default function HeroSection() {
+  const visitorData = useBotDResponse()
+
+  const [isLoading, setIsLoading] = useState(true)
```
Lisenish, line 35: This `isLoading` property should be returned from you custom `useBotDResponse` hook

src/hooks/useBotDResponse.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState } from 'react'
+import Botd from '@fpjs-incubator/botd-agent'
+import { BOTD_TOKEN } from '../constants/env'
+
+export const useBotDResponse = () => {
```
Lisenish, line 5: 1. I think you can call it just `useBotD`.
2. We use function declaration everywhere, no?

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -1,13 +1,83 @@
...
+
+  useEffect(() => {
+    setIsLoading(true)
+    if (visitorData) {
+      if (visitorData.bot.automationTool.probability > 0) {
+        setIsBot(true)
+        setdetectedBots((prevState) => ({
+          ...prevState,
+          automationTool: true,
+        }))
```
Lisenish, line 52: You always need to update to dependant values in state together, probably you can replace these two states with one `useReducer`

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -1,13 +1,83 @@
...
+      }
+      setIsLoading(false)
+    }
+  }, [visitorData])
+
+  const scrollToElementById = (id: string) => {
+    const section = document.querySelector(`#${id}`)
+    section && section.scrollIntoView({ behavior: 'smooth', block: 'start' })
+  }
```
Lisenish, line 79: Looks like reusable helper, don't we have something similar already? 
Please extract to commons, may be helpful to use somewhere else

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -32,9 +107,156 @@ export default function HeroSection() {
...
       </Section>
     </Container>
   )
 }
+
+interface CardsSectionProps extends DetectedBots {
+  isLoading?: boolean
+}
+function CardsSection({ automationTool, browserSpoofing, searchEngine, vm, isLoading }: CardsSectionProps) {
+  const repeatElement = (length, fn) => Array.from({ length }, (_, i) => fn(i))
```
Lisenish, line 130: Also looks like a function which may be useful somewhere else as well, also it's not clear why do you need to recreate it on each render

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -32,9 +107,156 @@ export default function HeroSection() {
...
+
+  return isLoading ? (
+    <div className={styles.cards}>
+      {repeatElement(4, (i) => (
+        <LoadingCard key={i} />
+      ))}
+    </div>
```
Lisenish, line 137: It would look better if you just do 
```
if (isLoading) {
  return <JSX>
}

return <JSX>
```

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -32,9 +107,156 @@ export default function HeroSection() {
...
+  ) : (
+    <div className={styles.cards}>
```
Lisenish, line 139: Maybe `section`?

Lisenish (commented):
Approved but there're some comments

Lisenish (approved):

Lisenish (changes requested):
Actually, I decided that I will take a 2nd look at the next version of the custom hook

src/styles/vendors/tippy.scss
natapg: I tried the tiptool in several browsers and it works fine, but I'm not sure why it was here this property, maybe I should find another way to solve the arrow problem without removing it

src/hooks/useBotDResponse.ts
natapg: thanks, I already sent a message to Rishat about this and he told me that he will talk with Valentin about this (and about who is going to create the docs in our docs page)

src/styles/_mixins.scss
```diff
@@ -310,3 +310,9 @@
     font-display: swap;
   }
 }
+
+@mixin scrollSection() {
+  @include media('>desktop') {
+    scroll-margin-top: 135px;
+  }
+}
```
Lisenish, line 318: Please add comment with explanation why do we need it, also maybe we should extract 135px to the constants?

Lisenish (approved):
Good job @natapg

PR #404, opened 2021-12-03
botd: API Response section

natapg:
Closes [WEB-287](https://fpjs.myjetbrains.com/youtrack/issue/WEB-287)


src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -107,15 +107,16 @@ export default function HeroSection() {
...
             <h2 className={styles.subTitle}>Bot detection in progress...</h2>
           ) : isBot ? (
             <h2 className={styles.subTitle}>You are a bot</h2>
           ) : (
             <h2 className={styles.subTitle}>You are not a bot</h2>
           )}
-          <p className={styles.seeDetails}>See details →</p>
+          <p className={styles.seeDetails} onClick={() => scrollToElementById('ApiResponseDetails')}>
+            See details →
+          </p>
```
tan-90, line 119: Does this work for you in Safari? For me it scrolls past the title and cuts a big part of the details section.

src/components/common/CodeWindowWithSelector/index.tsx
```diff
@@ -6,12 +6,20 @@ import { copyToClipboard } from '../../../helpers/clipboard'
 import { ReactComponent as CopySVG } from './CopySVG.svg'
 
 interface CodeWindowProps {
-  codeCDN: string
-  codeNPM: string
+  codeCDN?: string
+  codeNPM?: string
   hasLineNumbers?: boolean
+  singleCode?: string
```
tan-90, line 12: To avoid having a props that shouldn't exist together we could just change this component to receive an array of objects containing a tab name and code. That way if the array only has one element we show the simple code window, if it has more we show tabs.

tan-90 (changes requested):
No major issues but I want to take a look at the changes so I'm leaving it red for now.

src/components/common/CodeWindowWithSelector/index.tsx
natapg: good idea!

src/components/botd/HeroSection/HeroSection.tsx
natapg: yeah,for me too, I have to take a look, but also I'm waiting to have more sections to see how it behaves since in my case it doesn't scroll more because it reaches the end of the page

src/components/common/CodeWindowWithSelector/index.tsx
natapg: fixed @tan-90

src/components/botd/HeroSection/HeroSection.tsx
natapg: fixed @tan-90 , please check again! the header bar was hiding the element to be scrolled

tan-90: Works perfectly now 🎉

tan-90 (approved):

PR #411, opened 2021-12-06
Botd: Added Cloud Integration Section

natapg:


src/components/botd/IntegrationSection/IntegrationSection.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export default function IntegrationSection() {
+  return (
+    <Container className={styles.container}>
+      <TopIconSvg className={styles.topIcon} />
+      <h2 className={styles.title}>BotD integrates with all major cloud providers</h2>
+      <h3 className={styles.subTitle}>
+        Cloud integrations enable BotD to run at edge in a secure context to harden and protect your bot detection
+        logic. See for yourself - all cloud integrations are 100% open source.
+      </h3>
```
tan-90, line 21: I think I would make this a `p`, it feels more like a text body.

tan-90 (approved):
Just one minor comment we could discuss, everything else looks good.

PR #429, opened 2021-12-14
Botd: Related Articles section

natapg:
image
https://www.figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI-%2B?node-id=2836%3A35945

### DIFF
--- a/src/pages/botd/index.tsx
+++ b/src/pages/botd/index.tsx
@@ -2,6 +2,9 @@ import React from 'react'
 import { LayoutTemplate } from '../../components/Layout'
 import BreadcrumbsSEO from '../../components/Breadcrumbs/BreadcrumbsSEO'
 import { GeneratedPageContext } from '../../helpers/types'
+import Container from '../../components/common/Container'
+import RelatedArticles from '../../components/RelatedArticles/RelatedArticles'
+import { PostProps } from '../../components/Post/Post'
 
 import HeroSection from '../../components/botd/HeroSection/HeroSection'
 import GenerateKeySection from '../../components/botd/GenerateKeySection/GenerateKeySection'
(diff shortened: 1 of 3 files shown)


src/pages/botd/index.tsx
```diff
@@ -22,11 +27,19 @@ export default function Botd({ pageContext }: AccountSharingProps) {
   let siteMetadata = useSiteMetadata()
   siteMetadata = {
     ...siteMetadata,
-    title: 'TODO', //TODO
-    description: 'TODO', //TODO
+    title: 'TOpen Source JavaScript Bot Detection Library - BotDODO',
```
tan-90, line 30:
```suggestion
    title: 'Open Source JavaScript Bot Detection Library - Bot',
```

src/pages/botd/index.tsx
```diff
@@ -36,6 +49,9 @@ export default function Botd({ pageContext }: AccountSharingProps) {
       <DocumentationSection />
       <IntegrationSection />
       <FeaturesSection />
+      <Container size='large' className={styles.relatedArticles}>
+        <RelatedArticles article={content} count={4} title='BotD related articles' titleIsCentered limitPostLines />
```
tan-90, line 53: We've been needing the related articles section on pages a lot so it's probably a good idea to change it so we don't need to construct the entire `PostProps` object every time. We could do that with a `PartialWithRequired`:
```
type PartialWithRequired<T, K extends keyof T> = Pick<T, K> & Partial<T>

article: PartialWithRequired<PostProps, 'tags'>
```

tan-90 (approved):
Only minor comments, approving in advance but please take a look.

src/pages/botd/index.tsx
natapg: oh wow, how did I miss this

PR #433, opened 2021-12-15
BotD: added demo error state

natapg:
Closes [WEB-297](https://fpjs.myjetbrains.com/youtrack/issue/WEB-297)
errorStateDemo


src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -142,62 +151,69 @@ function CardsSection({ automationTool, browserSpoofing, searchEngine, vm, isLoa
...
-            <p>
-              <strong>Virtual machine detection</strong> is useful to detect click farms, automated review fraud and
-              junk content generation. It&apos;s a strong signal that improves the reliability and accuracy of the
-              previous three detectors.
-            </p>
-          }
-        />
+        {repeatElement(4, (i) => (
+          <LoadingCard error key={i} />
+        ))}
```
Lisenish, line 159: Can't we merge it with the upper condition?
```
if (isLoading || hasError) {
    <LoadingCard error={hasError} key={i} />
}
```

src/components/botd/HeroSection/HeroSection.tsx
```diff
@@ -123,7 +131,7 @@ export default function HeroSection() {
           <p className={styles.seeDetails} onClick={() => scrollToElementById('ApiResponseDetails')}>
             See details →
           </p>
-          <CardsSection {...botState} isLoading={isLoading} />
+          <CardsSection {...botState} hasError={hasError} isLoading={isLoading} />
```
Lisenish, line 134: I'm not sure but maybe it worth extracting to a local component because nested ternary operator not the best solution for the readability IMO. But if you think it's clear now and more convinient, I'm fine with it

src/hooks/useBotD.ts
```diff
@@ -27,10 +27,11 @@ export const useBotD = () => {
...
       setIsLoading(false)
+      setReload(false)
     }
 
     getVisitorData()
-  }, [])
+  }, [reload])
 
-  return { visitorData, isLoading, hasError, error }
+  return { visitorData, isLoading, hasError, error, setReload }
```
Lisenish, line 36: I suggested to expose high-level function here, like `refresh()` which will set the internal state here by itslef, you probably don't need to expose `setReload` itself to the outside

Lisenish (approved):
@natapg I've left some comments please take a lookm but nothing major so I approve in advance

PR #460, opened 2021-12-29
[Solutions Page] Add Bottom links section

natapg:


src/templates/solution-content-template.tsx
```diff
@@ -28,6 +32,7 @@ export interface TemplateProps {
   docsUrl: string
   body: string | React.ReactNode
   contentComponent?: React.FunctionComponent<{ content: string | React.ReactNode; className?: string }>
+  BottomLinks?: BottomLink[]
```
tan-90, line 35: Should be `bottomLinks`

src/templates/solution-content-template.tsx
```diff
@@ -82,6 +88,15 @@ export default function SolutionContentTemplate({
           </section>
           <section className={styles.bodySection}>
             <ContentComponent content={body} className={styles.content} />
+            {BottomLinks && (
+              <div className={styles.bottomLinks}>
+                {BottomLinks.map(({ text, url }) => (
+                  <a key={url} href={url}>
```
tan-90, line 94: We probably want `"target=_blank"`

tan-90 (approved):
Nothing major so I'm approving in advance but please take a look at the comments.

PR #462, opened 2021-12-30
feat: added ability to customize notification bar from cms

natapg:
Closes [WEB-316](https://fpjs.myjetbrains.com/youtrack/issue/WEB-316)


src/components/Layout/index.tsx
```diff
@@ -11,22 +11,57 @@ import { BASE_URL, URL } from '../../constants/content'
...
+  `)
 
-  return <LayoutTemplate siteMetadata={siteMetadata}>{children}</LayoutTemplate>
+  return data.notificationBarYaml?.showNotBar ? (
+    <LayoutTemplate
+      siteMetadata={siteMetadata}
+      headerBarTitle={<div dangerouslySetInnerHTML={{ __html: data.notificationBarYaml.barBody ?? '' }} />}
+      headerBarLinkUrl={data.notificationBarYaml.url}
+      headerBarArrowText={data.notificationBarYaml.arrowText}
+      headerBarColor={data.notificationBarYaml.backgroundColor}
```
tan-90, line 39: We have a lot of props for the header bar so it might be time to turn it into an object so we can just do `notificationBar={data.notificationBarYaml}`

src/components/Layout/index.tsx
```diff
@@ -72,7 +107,12 @@ export function LayoutTemplate({ children, siteMetadata }: LayoutTemplateProps)
         <link rel='preconnect' href={TLS_ENDPOINT} />
         <link rel='preconnect' href={FPJS_VISITORS_ENDPOINT} />
       </Helmet>
-      <Header />
+      <Header
+        headerBarTitle={headerBarTitle}
+        headerBarLinkUrl={headerBarLinkUrl}
+        headerBarArrowText={headerBarArrowText}
+        headerBarColor={headerBarColor}
```
tan-90, line 114: Could turn this into an object too.

tan-90 (approved):
Approving in advance but I left some minor comments, please take a look.

PR #491, opened 2022-01-20
fix: removed dates from case studies overview page

natapg:
Closes [WEB-323](https://fpjs.myjetbrains.com/youtrack/issue/WEB-323)


src/components/CaseStudy/CaseStudy.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <h1 className={styles.title}>{title}</h1>
+          <p className={styles.description}>{description}</p>
+        </div>
+      </div>
+    </Link>
+  )
+}
+
+// TODO [VL] Write a proper type for this. It's not as straight forward because it has to accept a lot of different generated gatsby types.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tan-90, line 28: Generated Gatsby types still scare me :sadge:

src/templates/case-studies.tsx
```diff
@@ -7,16 +7,18 @@ import { GeneratedPageContext } from '../helpers/types'
...
+import CaseStudy, { mapToCaseStudy } from '../components/CaseStudy/CaseStudy'
 
 interface CaseStudyProps {
   data: GatsbyTypes.CaseStudiesQuery
   pageContext: CaseStudiesContext
 }
 export default function CaseStudies({ data, pageContext }: CaseStudyProps) {
-  const { edges: posts } = data.posts
+  const { edges: caseStudiesData } = data.caseStudies
+  const caseStudies = caseStudiesData.map(({ node }) => node).map((node) => mapToCaseStudy(node))
```
tan-90, line 21: Is it possible to do this in a single map? I remember using two maps on the case studies but can't remember if there was a reason.

`const caseStudios = caseStudiesData.map(({node}) => mapToCaseStudy(node))`

tan-90 (approved):
Just one minor comment, please take a look but I'm approving in advance.

src/templates/case-studies.tsx
natapg: yeah, it is possible! I'll change it

PR #500, opened 2022-01-27
Authors page facelift

natapg:
Closes [WEB-328](https://fpjs.myjetbrains.com/youtrack/issue/WEB-328)

[Mockup](https://www.figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI-%2B?node-id=3278%3A46108)


src/components/AuthorSummary/AuthorSummary.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <h3 className={styles.title}>Role</h3>
+        <span className={styles.description}>{role}</span>
+        {bio && (
+          <>
+            <h3 className={styles.bioTitle}>Bio</h3>
+            <span className={styles.description}>{bio}</span>
+          </>
+        )}
+      </div>
+      <h3 className={styles.articlesTitle}>{`${author} Articles`}</h3>
```
tan-90, line 32: I'd change it to `${author}'s Articles` but probably better to talk to a native speaker about that.

src/templates/author-tag.tsx
```diff
@@ -0,0 +1,109 @@
...
+    description: `We are an open source powered company working to prevent online fraud for websites of all sizes. Read our articles written by ${author} on our blog.`,
+    siteUrl: `${siteMetadata.siteUrl}${pathname}`,
+  }
+
+  return (
+    <LayoutTemplate siteMetadata={siteMetadata}>
+      {breadcrumbs && (
+        <Container size='large'>
+          <BreadcrumbsSEO breadcrumbs={breadcrumbs} />
+        </Container>
```
tan-90, line 43: If we're only using the SEO there's no need to have a container.

src/templates/author.tsx
```diff
@@ -40,36 +40,20 @@ export default function Author({ data, pageContext }: AuthorProps) {
       {breadcrumbs && (
         <Container size='large'>
           <BreadcrumbsSEO breadcrumbs={breadcrumbs} />
-          <Breadcrumbs breadcrumbs={breadcrumbs.slice(1)} />
         </Container>
```
tan-90, line 43: We can remove the container here as well.

src/templates/author.tsx
```diff
@@ -40,36 +40,20 @@ export default function Author({ data, pageContext }: AuthorProps) {
...
-          </div>
-        </div>
-
+        <AuthorSummary author={author} role={role} bio={bio} photo={photo} />
         <Posts
           name={`${author} Articles`}
           posts={posts
             .map(({ node }) => node)
             .map((node) => mapToPost(node))
             .map((post) => ({ ...post }))}
```
tan-90, line 53: Why do we need `{ ...post }`? Is it because of typing? Could also change this to use a single map.

src/components/AuthorSummary/AuthorSummary.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Link } from 'gatsby'
+
+interface AuthorSummaryProps {
+  author: string
+  role?: string
+  photo?: IGatsbyImageData
+  bio?: string
+}
+export default function AuthorSummary({ author, photo, role, bio }: AuthorSummaryProps) {
+  return (
```
tan-90, line 14: Since it's only used on the author's page, does it need to be clickable? It would only remove the active tags right?

tan-90 (changes requested):
Pointed out a couple minor things, requesting changes so we can also discuss some of the questions.

src/components/AuthorSummary/AuthorSummary.tsx
natapg: yeah, that's the idea since we don't have breadcrum here, but I can make it clickable only on the tag page I think

src/components/AuthorSummary/AuthorSummary.tsx
natapg: good point, I've asked Savannah on our internal channel

src/templates/author.tsx
natapg: My bad, I think I had a typing problem at first, but with a single map it's fine, I fixed it

src/components/AuthorSummary/AuthorSummary.tsx
natapg: @tan-90 Savannah's reply -> yes ’s is right!

tan-90 (approved):

PR #511, opened 2022-02-03
Integrated chromatic with storybook

natapg:
Closes [WEB-87](https://fpjs.myjetbrains.com/youtrack/issue/WEB-87)


.github/workflows/ci.yml
```diff
@@ -147,3 +147,29 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # Required to retrieve git history
+          fetch-depth: 0
+
+      - name: yarn install
+        uses: bahmutov/npm-install@v1
+        with:
+          install-command: yarn --frozen-lockfile
```
Lisenish, line 163: Why do we use 3rd party action for install btw? I see that it's same above but can't we do it in the same way as we have it in Dashboard (just direct yarn command with github actions cache)

.github/workflows/ci.yml
```diff
@@ -147,3 +147,29 @@ jobs:
...
+
+      - name: yarn build
+        run: yarn run build
+        env:
+          GATSBY_EXPERIMENTAL_PAGE_BUILD_ON_DATA_CHANGES: true
```
Lisenish, line 168: Are you sure we need to build whole website for this? We don't have it in Dashboard and I think chromatic action should build storybook by itself and we probably shouldn't need website artifacts here.

In case if we need it, maybe it makes sense to share build artifacts with the other jobs instead of building it again from the scratch?

.storybook/main.js
```diff
@@ -39,11 +39,22 @@ module.exports = {
       propFilter: (prop) => (prop.parent ? !/node_modules/.test(prop.parent.fileName) : true),
     },
   },
+  babel: async (options) => {
```
Lisenish, line 42: Please leave a comment why do we need it otherwise In 6 months nobody would be able to tell which settings is needed for what and can we change them or not.

Lisenish (changes requested):

.github/workflows/ci.yml
natapg: we can do it like in the dashboard, I was using this action to be consistent with the other steps, I will change it

.github/workflows/ci.yml
natapg: Yeah, we need to build first
`babel-preset-gatsby has been loaded, which consumes config generated by the Gatsby CLI ... run `gatsby build` first.`
but makes sense, I didn't know that artifacts can be shared between jobs, good idea

.github/workflows/ci.yml
natapg: done @Lisenish

.github/workflows/ci.yml
natapg: done @Lisenish

.github/workflows/ci.yml
```diff
@@ -53,46 +64,59 @@ jobs:
...
-
-      - name: Cache gatsby public folder
-        uses: actions/cache@v2
-        id: gatsby-public-folder
-        with:
-          path: public/
-          key: ${{ runner.os }}-public-gatsby
-          restore-keys: |
-            ${{ runner.os }}-public-gatsby
+        uses: actions/checkout@v2.4.0
```
Lisenish, line 90: Why do we need a specific version here? Maybe we can leave just major version then it will use the latest one automatically. I see the same in different places as well, please change all the versions to the major ones everywhere

.github/workflows/ci.yml
```diff
@@ -53,46 +64,59 @@ jobs:
...
+      - name: upload build
+        uses: actions/upload-artifact@v2.3.1
+        with:
+          name: build
+          path: build.tar.gz
+      - name: upload cache
+        uses: actions/upload-artifact@v2.3.1
+        with:
+          name: cache
+          path: cache.tar.gz
```
Lisenish, line 81: It confused me at first since I didn't get why do we need to upload cache manually since `actions/cache` already doing it automatically but then I figured out that here we're uploading gatsby's cache and not yarn's cache, let's maybe specify that in the name so it's clear what kind of cache it is

Lisenish (approved):

PR #518, opened 2022-02-10
fix: added phone number to contact sales form

natapg:
Closes [WEB-331](https://fpjs.myjetbrains.com/youtrack/issue/WEB-331)


package.json
```diff
@@ -70,6 +70,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-helmet": "^6.0.0",
+    "react-phone-input-2": "^2.15.0",
```
Lisenish, line 73: Let's pin it, this library is pretty small and can't be trusted so we need to check every update carefully.

src/components/ContactSalesForm/index.tsx
```diff
@@ -64,6 +76,12 @@ export default function ContactSalesForm() {
     }
   }
 
+  const addPhone = () => {
```
Lisenish, line 79: The usual name convention for even handlers to start them with prefix `handle...` + the name of the event.
So here it should be something like `handlePhoneChange`

src/components/ContactSalesForm/index.tsx
```diff
@@ -122,6 +140,18 @@ export default function ContactSalesForm() {
...
               />
+              <label className={styles.label} htmlFor='phone'>
+                Phone number
+              </label>
+              <PhoneInput
+                inputClass={styles.phoneInput}
+                buttonClass={styles.phoneDropdown}
+                country={'us'}
+                value={phone}
+                onChange={() => addPhone()}
```
Lisenish, line 151: Why not just ?

```suggestion
                onChange={addPhone}
```

src/components/ContactSalesForm/index.tsx
```diff
@@ -64,6 +76,12 @@ export default function ContactSalesForm() {
     }
   }
 
+  const addPhone = () => {
+    const inputElement = document.getElementsByName('phoneInput')[0] as HTMLInputElement
```
Lisenish, line 80: You very rarely need to use raw DOM selectors in React. You need to use `useRef` hook for this and pass it to `inputProps.ref`

src/components/ContactSalesForm/index.tsx
```diff
@@ -64,6 +76,12 @@ export default function ContactSalesForm() {
...
+    setPhone(inputElement.value)
```
Lisenish, line 81: Why're are you using `inputElement.value` here? You should get the phone value from the  handler function arguments

Lisenish (changes requested):

src/components/ContactSalesForm/index.tsx
natapg: I asked Maciej about this, if he needs the number to be formatted or not (like +1 125 2131 1231 vs +112521311231)
in case he needs it formatted, this way I can get it with the format, if he doesn't need it I can get it from the handler function arguments

src/components/ContactSalesForm/index.tsx
natapg: yeah I know, the issue is that I don't have direct access to the input element and that's why I decided to do it this way, but I'll see if I can do it with the inputProps parameter

natapg: It's possible thanks to this fix :) changing it https://github.com/bl00mber/react-phone-input-2/issues/187

src/components/ContactSalesForm/index.tsx
```diff
@@ -148,9 +149,9 @@ export default function ContactSalesForm() {
                 buttonClass={styles.phoneDropdown}
                 country={'us'}
                 value={phone}
-                onChange={() => addPhone()}
+                onChange={(phone) => handlePhoneChange(phone)}
```
Lisenish, line 152: nit: it can be just this, since signature is the sasme:

```suggestion
                onChange={handlePhoneChange}
```

Lisenish (approved):

PR #523, opened 2022-02-14
fix: small fixes to the FAQ section

natapg:
Closes [WEB-343](https://fpjs.myjetbrains.com/youtrack/issue/WEB-343)

I'm changing toLocaleString() by a js function because for some reason it's not working on prod.
The width of the content is taken from the [mockup](https://www.figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI-%2B?node-id=1483%3A11086)


src/pages/pricing/index.tsx
```diff
@@ -69,7 +70,7 @@ function FAQSection() {
       answer: (
         <p className={styles.faqContent}>
           Customers are billed on a monthly basis based on API requests made over the billing period. The minimum paid
-          plan is ${(overagePrice / 100) * prepaidQuantity}/mo for {prepaidQuantity.toLocaleString()} API requests - any
+          plan is ${(overagePrice / 100) * prepaidQuantity}/mo for {numFormatter(prepaidQuantity)} API requests - any
```
borislobanov92, line 73: There is no need for a custom formatter, it will be cleaner if we can make `toLocaleString()` work. The reason it's not working is because `prepaidQuantity` is actually a string and not a number.

The simplest fix, which I don't particularly like, is to do `Number(prepaidQuantity).toLocaleString()`. The reason I don't like it is because you'd have to do it everywhere you are using this value.

The good fix I'd love to see here is to figure out at what point the number returned by the original endpoint turns into a string and prevent that. It may not be worth it just for this particular fix since it might take a lot of time but it would probably be a good opportunity to investigate and learn more about how Gatsby works under the hood (I don't know what causes the problem). If this investigation takes more than 2 hours, I'd just do simple conversion to a number and forget about it for now.

borislobanov92 (changes requested):

src/pages/pricing/index.tsx
borislobanov92: https://stackoverflow.com/questions/49300485/tolocalestring-is-not-working

natapg: good catch, thanks! you are right, the problem is that the query returns the values in string type, so I added the conversion to the custom hook that performs the query

borislobanov92 (changes requested):

src/pages/pricing/index.tsx
```diff
@@ -94,7 +94,7 @@ function FAQSection() {
       question: 'I am a developer. Can I use FingerprintJS Pro for free?',
       answer: (
         <p className={styles.faqContent}>
-          Yes - developers and small sites can use FingerprintJS Pro for free forever, with up to{' '}
+          Yes you can. Developers and small sites can use FingerprintJS Pro for free forever, with up to{' '}
```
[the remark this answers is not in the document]
natapg: I've asked Savannah about this, which one should we use, but she hasn't answered me yet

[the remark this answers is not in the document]
natapg: Savannah prefers hyphenation, so I made the changes

borislobanov92 (approved):

PR #539, opened 2022-03-02
BNPL : Added first 3 sections

natapg:
Ticket: https://fpjs.myjetbrains.com/youtrack/issue/WEB-342

Mockup: https://www.figma.com/file/niHrIRowTbqQhDO7W5CPNf/FPJS-WEB-UI-%2B?node-id=3856%3A50732

https://deploy-preview-539--fingerprintjscom.netlify.app/buy-now-pay-later/


src/components/accountSharing/ReadStudyCaseSection/ChooseAnswerSVG.svg
```diff
@@ -1,74 +1,61 @@
...
-        <path d="M320 0V440" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M400 0V440" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M480 0V440" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M560 0V440" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M600 60H0" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M600 140H0" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M600 220H0" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M600 300H0" stroke="#E3E3ED" stroke-width="1.2" />
-        <path d="M600 380H0" stroke="#E3E3ED" stroke-width="1.2" />
+<svg fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 608 440">
```
natapg, line 1: I noticed that this SVG was not displaying properly so I fixed it

src/components/AlternatingImagesText/AlternatingImagesText.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export interface AlternatingImagesTextProps {
+  blocks: BlockWithImage[]
+}
+export default function AlternatingImagesText({ blocks }: AlternatingImagesTextProps) {
+  return (
+    <>
+      {blocks.map((block, index) => (
+        <Container size='large' className={styles.container} key={index}>
```
tan-90, line 19: It won't be an issue in this case because the content is static but for dynamic content having the key be an index wouldn't work so I would change this to be the title instead.

src/components/AlternatingImagesText/AlternatingImagesText.tsx
```diff
@@ -0,0 +1,29 @@
...
+  content: React.ReactNode
+}
+
+export interface AlternatingImagesTextProps {
+  blocks: BlockWithImage[]
+}
+export default function AlternatingImagesText({ blocks }: AlternatingImagesTextProps) {
+  return (
+    <>
+      {blocks.map((block, index) => (
```
tan-90, line 18: Optional: `({ image: Image, title, content }, index)`

src/pages/buy-now-pay-later/Bnpl.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  padding-bottom: rem(64px);
+  p {
+    font-size: rem(18px) !important;
+    font-weight: 500;
+    color: v('brand-water') !important;
+
+    @include media('<=phoneLandscape') {
+      font-size: rem(16px);
+    }
+  }
```
tan-90, line 11: Can't we replace this with a named class to get rid of the `!important`? We shouldn't have to use it unless it's for markdown content.

tan-90 (changes requested):
Nothing major but I just wanna take another look.

src/components/AlternatingImagesText/AlternatingImagesText.tsx
natapg: since I don't have an ID take the index, as suggested in the documentation, but yeah, I imagine that there should not be equal titles either
https://reactjs.org/docs/lists-and-keys.html#keys

tan-90: I actually didn't know the docs suggested the index as a last option, thanks for the link!

PR #543, opened 2022-03-04
[WIP] Migrate gatsby config files to TS

natapg:
Closes [WEB-351](https://fpjs.myjetbrains.com/youtrack/issue/WEB-351)


gatsby-node.ts
```diff
@@ -37,17 +57,16 @@ async function getFolderEdges(folder, graphql, filter = '') {
...
+    throw new Error(allMarkdown.errors)
   }
-
-  return data.allMarkdownRemark.edges
+  return allMarkdown?.data?.allMarkdownRemark.edges
 }
 
-async function getArrayFieldValues(graphql, name) {
-  const { data, errors } = await graphql(`
+const getArrayFieldValues = async (graphql: CreatePagesArgs['graphql'], name: string) => {
```
tan-90, line 68: Minor: why change to an arrow function? On the dashboard we usually prefer function declarations.

gatsby-node.ts
```diff
@@ -138,55 +157,67 @@ exports.createPages = async ({ actions, graphql }) => {
   const { createPage } = actions
 
   const pages = await getFolderEdges('index', graphql)
-  pages.forEach((edge) => createPageFromEdge(edge, createPage))
+  if (pages) {
+    pages.forEach((edge) => createPageFromEdge(edge, createPage))
+  }
```
tan-90, line 162: Can we change it to this on all places instead of adding if statements?

```suggestion
 pages?.forEach((edge) => createPageFromEdge(edge, createPage))
```

gatsby-node.ts
```diff
@@ -269,20 +300,42 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
+      typeof node.fileAbsolutePath === 'string' &&
+      /(blog)\/.*.md$/.test(node.fileAbsolutePath)
   )
   const authors = getNodes().filter(
-    (node) => node.internal.type === 'MarkdownRemark' && /(author)\/.*.md$/.test(node.fileAbsolutePath)
+    (node) =>
+      node.internal.type === 'MarkdownRemark' &&
+      typeof node.fileAbsolutePath === 'string' &&
+      /(author)\/.*.md$/.test(node.fileAbsolutePath)
   )
```
tan-90, line 332: If we move the `as AuthorNode[]` to here it will already have the correct type and there will be no need to cast before using it.

gatsby-node.ts
```diff
@@ -269,20 +300,42 @@ exports.onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
...
 
   blogPosts.forEach((node) => {
-    if (node.frontmatter.authors) {
-      const authorNodes = authors.filter((otherNode) => node.frontmatter.authors.includes(otherNode.frontmatter.title))
-
+    const blogNode = node.frontmatter as BlogNode['frontmatter']
+    if (blogNode.authors) {
+      const authorsNode = authors as unknown as AuthorNode[]
```
tan-90, line 337: Doing the change in the previous comment we could get rid of this and just use `authors` instead.

tan-90 (changes requested):

gatsby-config.ts
```diff
@@ -86,22 +87,22 @@ module.exports = {
...
         name: 'index',
       },
     },
     {
       resolve: 'gatsby-source-filesystem',
       options: {
-        path: `${__dirname}/src/img`,
+        path: path.resolve(`src/img`),
         name: 'images',
       },
```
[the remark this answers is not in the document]
natapg: I think there is no reason, I will change it

gatsby-node.ts
natapg: I just followed the Gatsby migration guide which uses arrow functions, but I can change it

natapg:
@tan-90 I've made the changes that you requeted! thanks!

tan-90 (approved):

PR #560, opened 2022-03-23
feat: integrated turing into contact sales form

natapg:


src/constants/content.ts
```diff
@@ -34,10 +34,12 @@ export const URL = {
   signupUrl: 'https://dashboard.fingerprintjs.com/signup',
   statusUrl: 'https://status.fingerprintjs.com',
   supportMail: 'support@fingerprintjs.com',
+  salesMail: 'sales@fingerprintjs.com',
   discordServerURL: 'https://discord.gg/ad6R2ttHVX',
 } as const
 
 export const MAILTO = { mailToUrl: `mailto:${URL.supportMail}` } as const
+export const MAILTOSALES = { mailToUrl: `mailto:${URL.salesMail}` } as const
```
tan-90, line 42: Minor: `MAILTO_SALES`

src/components/ContactSalesForm/index.tsx
```diff
@@ -27,53 +27,61 @@ export default function ContactSalesForm() {
...
   const { formState, errorMessage, updateFormState, updateErrorMessage } = useForm(Forms.ContactSales)
   const { landingPage, previousPage, utmParams } = useViewTracking()
-  useBotD(BOTD_PUBLIC_TOKEN_CONTACT_SALES, BOTD_SECRET_TOKEN_CONTACT_SALES)
 
   async function handleSubmit(e) {
     e.preventDefault()
 
-    updateFormState(FormState.Loading)
+    try {
+      await Turing.ready(async () => {
```
tan-90, line 35: Is it possible to call `Turing.ready` ahead of time so when we want to submit the only need to call `Turing.execute`? (similar to how we load FPJS ahead of time on the dashboard)

src/components/ContactSalesForm/index.tsx
```diff
@@ -27,53 +27,61 @@ export default function ContactSalesForm() {
...
+        const sessionId = await Turing.execute()
+        updateFormState(FormState.Loading)
```
tan-90, line 37: We should set the loading state before starting to get the sessionId.

src/components/botd/GenerateKeySection/GenerateKeySection.tsx
```diff
@@ -41,9 +43,12 @@ export default function GenerateKeySection({ requestId }: GenerateKeySectionProp
   async function handleSubmit(e) {
     e.preventDefault()
 
-    updateFormState(FormState.Loading)
-
-    function onError() {
+    function onError(error?) {
+      if (error?.error?.message === 'turing challenge is not verified') {
```
tan-90, line 47: In this case, won't `error.param` be `Turing`? If possible we shouldn't rely on error messages for any logic.

tan-90 (changes requested):

src/components/ContactSalesForm/index.tsx
natapg: the problem doing that is, If you close the Turing modal, the button remains in the "loading" state, and I can't think of how to solve that in another way

src/components/ContactSalesForm/index.tsx
natapg: fixed! please check if the implementation is ok @tan-90

src/components/botd/GenerateKeySection/GenerateKeySection.tsx
```diff
@@ -109,6 +108,7 @@ export default function GenerateKeySection({ requestId }: GenerateKeySectionProp
         return
       }
     }
+    initTuring()
```
tan-90, line 111: This way we would still need to init it every time the form is submitted. I was thinking about doing something like we have on the dashboard.

We could have a helper with a const variable for the initialized Turing instance and then either return the existing one or initialize it if it doesn't exist.


For reference you could take a look at `src/helpers/vendor/fpjs.ts` on the dashboard.

tan-90 (approved):

PR #578, opened 2022-04-08
fix: two calls to fpjs in the home page

natapg:


package.json
```diff
@@ -80,7 +80,7 @@
     "swiper": "^8.0.3"
   },
   "devDependencies": {
-    "@babel/core": "7.17.8",
```
natapg, line 83: this small update generates a lot of warnings so i decided to downgrade to the previous version, I will investigate the problem later

tan-90 (approved):

PR #587, opened 2022-04-13
Fix: added visits loading state

natapg:
Closes [WEB-403](https://fpjs.myjetbrains.com/youtrack/issue/WEB-403)


src/components/FpjsWidget/FpjsWidget.module.scss
```diff
@@ -22,18 +22,11 @@
   width: 100%;
 
   svg {
-    opacity: 0;
     outline: none;
     width: 2rem;
     height: 2rem;
   }
 
-  &.loaded {
```
natapg, line 31: with the loading status it is not necessary to hide the tooltip icon

src/components/FpjsWidget/index.tsx
```diff
@@ -27,7 +27,7 @@ export default memo(function FpjsWidget() {
   const [isLoading, setIsLoading] = useState(true)
   const [isLoaded, setIsLoaded] = useState(false)
   const [hasError, setHasError] = useState(false)
-  const { getData, error } = useVisitorData(getConfig, { immediate: false })
+  const { getData, error, isLoading: fjpsIsLoading } = useVisitorData(getConfig, { immediate: false })
```
Lisenish, line 30: We can rename `error` here to `indeficationError`.  And `hasError` to something like `historyLoadError`
Then we can declare a variable `const hasError = indeficationError || historyLoadError`

In this case we won't need this code below in the effect
```    
if (error) {
      setHasError(true)
      return
    }
```

Lisenish, line 30: We can rename `error` here to `indeficationError`.  And `hasError` to something like `historyLoadError`
Then we can declare a variable `const hasError = indeficationError || historyLoadError`

In this case we won't need this code below in the effect
```    
if (error) {
      setHasError(true)
      return
    }
```

Lisenish, line 30: typo, should be`**fpjs**IsLoading`

Lisenish, line 30: typo, should be`**fpjs**IsLoading`

src/components/FpjsWidget/index.tsx
```diff
@@ -212,7 +234,7 @@ interface LoadingStateProps {
   hasError: boolean
 }
 
-function LoadingState({ hasError }: LoadingStateProps) {
+function FullLoadingState({ hasError }: LoadingStateProps) {
```
Lisenish, line 237: Why is is called `Full...`? Maybe `DetailsLoadingState`/ `VisitDetailsLoadingState`/`VisitDetailsSkeleton` would be better?

Lisenish: Ok figured it out, I think you incorrectly composed component, we shouldn't manually compose all the possible states, we should rather have two components: one for histroy and the other for details part and they both should have their own loading state. 
It should be something like this:

```html
<div className={styles.container}>
  <VistHistory isLoading={isVisitHistoryLoading} ... />
  <VistDetails isLoading={isVisitDetailsLoading} ... />
</div>
```

So in this way you don't need to manually crate combinations fro the state when you have only visits loading, or both, or only details

Lisenish: In general it should look something like this:

```jsx 
if (hasError) {
 return <FpjsWidgetLoaded hasError />
}

if (isVisitHistoryLoading || isVisitDetailsLoading) {
  return <LoadingState isVisitDetailsLoading={isVisitDetailsLoading} isVisitDetailsLoading={isVisitDetailsLoading}  />

return <FpjsWidgetLoaded ... />
  
}
```

Lisenish (changes requested):
@natapg I've left some comments please take a look.
I also noticed that we always render both mobile and desktop versions, could you please remind me, it's because of the SSR? I'm not sure if it's the best decision, we probably need to discuss and potentially create a separate ticket for that

natapg:
thanks @Lisenish , I think I got your idea, please check :)

src/components/FpjsWidget/index.tsx
```diff
@@ -26,21 +25,25 @@ export default memo(function FpjsWidget() {
...
   const [isLoading, setIsLoading] = useState(true)
   const [isLoaded, setIsLoaded] = useState(false)
-  const [hasError, setHasError] = useState(false)
-  const { getData, error, isLoading: fjpsIsLoading } = useVisitorData(getConfig, { immediate: false })
+
   const rollbar = useRollbar()
 
+  const {
+    getData,
+    error: indeficationError,
```
Lisenish, line 33: nit: `indefication` is not a word

Lisenish (approved):
@natapg Good job, much cleaner 👍 Only one nitpick

PR #596, opened 2022-04-21
fix: show current visit when available on demo page

natapg:
Closes [WEB-403](https://fpjs.myjetbrains.com/youtrack/issue/WEB-403)


src/components/demo/DemoSection/DemoSection.tsx
```diff
@@ -27,6 +27,21 @@ export default function DemoSection() {
       if (!visitorId) {
         const data = await getData(true)
         setVisitorId(data?.visitorId)
+        const visitsFpjs = [
+          {
+            ...data,
+            timestamp: 1,
```
borislobanov92, line 33: can we take the timestamp from the data? it would probably be more fair

borislobanov92 (approved):

src/components/demo/DemoSection/VisitsSection/VisitsSection.tsx
```diff
@@ -76,9 +76,15 @@ export default function VisitsSection({ isLoading, currentVisit, visits }: Visit
             </tr>
             <tr className={styles.selected}>
               <td>Current visit</td>
-              <td>
-                <Skeleton className={styles.visitSkeleton} width={24} />
-              </td>
+              {currentVisit ? (
+                <td className={classNames({ [styles.incognito]: currentVisit.incognito })}>
+                  {currentVisit.incognito ? 'Yes' : 'No'}
```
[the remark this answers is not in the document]
natapg: to show the current visit if we have it while we continue to show the skeleton in the following rows

src/components/demo/DemoSection/DemoSection.tsx
natapg: I can remove it, we are not using

PR #601, opened 2022-04-29
Fix: optimized images

natapg:
Removed some unused images (leftover from static Account Sharing page)
Change width to 1800px for large images
Compressed images without losing quality
Disabled generation of WEBP, AVIF

decreased build time from 15/16 minutes to 9/10 minutes


renovate.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": ["config:base", "schedule:weekly"],
-  "assignees": ["natapg", "Lisenish"],
+  "assignees": ["natapg"],
```
tan-90, line 3: 😢

renovate.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": ["config:base", "schedule:weekly"],
-  "assignees": ["natapg", "Lisenish"],
```
tan-90, line 3: 😢

tan-90 (approved):

PR #616, opened 2022-05-16
feat: change fingerprintjs to fingerprint

natapg:


static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: feat/fingerprintjs-to-fingerprint
```
tan-90, line 3: For old times sake: reminder to change back

tan-90 (approved):
Not reasonable to look through every single change but I took an overall look and did a couple of searches, it looks good so I'm approving.

The only thing I found was a `@fingerprintjs` on the OSS email, not sure if it's intentional, please take a look at the comment.

PR #622, opened 2022-05-25
fix: added amplitude to website

natapg:
[Closes WEB-424](https://fpjs.myjetbrains.com/youtrack/issue/WEB-424)

*Before merging add variables to new netlify site (fingerprint .com)*


src/helpers/amplitude.ts
```diff
@@ -0,0 +1,16 @@
+import amplitude from 'amplitude-js'
+import { AMPLITUDE_API_KEY } from '../constants/env'
+
+const amplitudeInit = () => {
+  amplitude.getInstance().init(AMPLITUDE_API_KEY)
+}
```
borislobanov92, line 6: Let's add an env variable for the api endpoint and pass the config `{ apiEndpoint: AMPLITUDE_API_ENDPOINT }` to the `init` method (the same way it's done in the dashboard). It should work with the default endpoint if the variable is not set and send events to a custom endpoint if we provide it.

PR #661, opened 2022-07-07
added animated accuracy section

natapg:
closes [WEB-391](https://fpjs.myjetbrains.com/youtrack/issue/WEB-391)


src/components/AccuracySection/DiagramSection/DiagramSection.tsx
```diff
@@ -34,11 +36,28 @@ export default function DiagramSection() {
     }
   }, [isActive, timeoutId])
 
+  useEffect(() => {
+    if (currentTab === tabOptions.AfterPro) {
+      /* eslint-disable  @typescript-eslint/no-explicit-any */
+      const background = document.getElementById('backgroundafter') as any
+      const text = document.getElementById('ebJ3ZqH7pWA1') as any
```
tan-90, line 43: I assume we need this because the player library doesn't have TS typings?

tan-90 (approved):
Looks good to me, just wanted to double check we're only using `any` if the package doesn't have typings.

PR #686, opened 2022-08-10
updated gatsby

natapg:
Changes:

1. Updated to Gatsby 4.20
2. Removed gatsby-plugin-typegen to use the feature [GraphQL Typegen](https://www.gatsbyjs.com/docs/how-to/local-development/graphql-typegen/)
3. Removed react-helmet to use the new [API Gatsby Head](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-head/) and [Gatsby Script](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-script/) to load scripts


gatsby-node.js
```diff
@@ -389,6 +389,8 @@ exports.createSchemaCustomization = ({ actions }) => {
       inlineCta: InlineCta
       bio: String @md
       bottomLinks: [BottomLink]
+      invertContent: Boolean!
```
natapg, line 392: https://www.gatsbyjs.com/docs/how-to/local-development/graphql-typegen/#non-nullable-types
these fields cannot be null

src/pages/404/index.tsx
```diff
@@ -25,3 +25,5 @@ const NotFoundPage = () => (
 )
 
 export default NotFoundPage
+
+export const Head: HeadFC = () => <title>Fingerprint Pro - Not found</title>
```
tan-90, line 29: Let's use a function declaration instead for consistency.

tan-90 (approved):
All looks good to me. Only some minor comments, please take a look.

Since this changes the way we set SEO tags we should double check the stuff that's important to marketing is still working but you probably did that already :bongo_blog:

static/admin/config.yml
```diff
@@ -1,6 +1,6 @@
 backend:
   name: git-gateway
-  branch: master
+  branch: gatsby/v4.20
```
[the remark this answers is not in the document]
tan-90: image

PR #702, opened 2022-08-22
fix: added botd and visitorid to amplitude call

natapg:


src/helpers/amplitude.ts
```diff
@@ -5,12 +5,15 @@ const amplitudeInit = () => {
   amplitude.getInstance().init(AMPLITUDE_API_KEY, undefined, { apiEndpoint: AMPLITUDE_API_ENDPOINT })
 }
 
-export const amplitudeLogEvent = (eventName: string, properties?: Record<string, string>) => {
+export const amplitudeLogEvent = (eventName: string, properties?: Record<string, string>, visitorId?: string) => {
   if (process.env.NODE_ENV !== 'development') {
     amplitudeInit()
+    if (visitorId) {
+      amplitude.setUserId(visitorId)
```
dcorso21, line 12: I don't think we want to set the visitorId as the userId. Rather, we want to store it as a property.

src/helpers/amplitude.ts
```diff
@@ -5,12 +5,15 @@ const amplitudeInit = () => {
...
+    }
     amplitude.getInstance().logEvent(eventName, properties)
   } else {
     // eslint-disable-next-line no-console
-    console.log(`Amplitude event ${eventName} triggered with props:`, properties)
+    console.log(`Amplitude event ${eventName}, visitorId: ${visitorId}, triggered with props:`, properties)
```
dcorso21, line 17: I'm curious why we are logging out these events (I realize it was here before).

dcorso21 (changes requested):
Main thing to address is the visitorId set as the userId IMO.

src/components/Layout/index.tsx
```diff
@@ -56,8 +59,14 @@ export function LayoutTemplate({ children, notificationBar }: LayoutTemplateProp
   }, [isEuUser, userCountry])
 
   useEffect(() => {
-    amplitudeLogEvent('view marketing page', { route: pathname })
-  }, [pathname])
+    if (visitorData && visitorId) {
+      amplitudeLogEvent(
+        'view marketing page',
+        { route: pathname, is_detected: visitorData.products.botd.data.bot.result },
```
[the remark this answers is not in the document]
jackspirou: @dcorso21 my understanding is @natapg was worried about using using "bot" incase bots are looking for the term "bot".  @natapg I thought we agreed on a different term though like `isAutomated` or something.

src/helpers/amplitude.ts
jackspirou: Yeah this is probably leftover from debugging

natapg: It is only for development (when `process.env.NODE_ENV === 'development'`)

src/helpers/amplitude.ts
natapg: changed, please check @dcorso21

src/helpers/amplitude.ts
```diff
@@ -5,12 +5,15 @@ const amplitudeInit = () => {
   amplitude.getInstance().init(AMPLITUDE_API_KEY, undefined, { apiEndpoint: AMPLITUDE_API_ENDPOINT })
 }
 
-export const amplitudeLogEvent = (eventName: string, properties?: Record<string, string>) => {
+export const amplitudeLogEvent = (eventName: string, properties?: Record<string, string>, visitorId?: string) => {
   if (process.env.NODE_ENV !== 'development') {
     amplitudeInit()
+    if (visitorId) {
+      amplitude.getInstance().identify(new amplitude.Identify().set('visitorId', visitorId))
```
dcorso21, line 12: After checking the docs, I see that this is valid. Optional: this is also valid: 

`amplitude.getInstance().setUserProperties({ visitorId });`

dcorso21 (approved):
Thanks, one optional change but looks good.

PR #706, opened 2022-08-28
fix: decrease bundle size

natapg:


tan-90 (approved):
`src/img/bg-full.png` survived for so long, I remember it being there when we first joined.

PR #707, opened 2022-08-29
fix: use botd and identification same sub

natapg:


tan-90 (approved):
Screen Shot 2022-08-30 at 11 56 32
Safari strikes again (some parts of the automation tool svg don't get rendered for some reason) 🥲

All else looks good though, approving in advance.

PR #712, opened 2022-09-01
fix: make one call to core api - update fingerprintjs-pro-react 

natapg:


src/components/FpjsWidget/index.tsx
```diff
@@ -42,7 +42,7 @@ export default memo(function FpjsWidget() {
 
     async function fetchVisits() {
       if (!visitorId) {
-        const data = await getData(true)
+        const data = await getData({ ignoreCache: true })
```
tan-90, line 45: I noticed that our `.editorconfig` has a `trim_trailing_whitespace` setting, I wonder if prettier on its own doesn't catch this. If so you might have to set the editor to format on save too.

PR #718, opened 2022-09-07
fix: added faq page

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/705


src/pages/resources/frequently-asked-questions-faqs/index.tsx
```diff
@@ -0,0 +1,554 @@
...
+}
+export default function FaqPage({ pageContext }: FaqPageProps) {
+  const breadcrumbs = pageContext.breadcrumb.crumbs
+
+  return (
+    <LayoutTemplate>
+      {breadcrumbs && <BreadcrumbsSEO breadcrumbs={breadcrumbs} />}
+      <Section className={styles.section}>
+        <Container size='large' className={styles.container}>
+          <header className={styles.header}>
```
tan-90, line 28: I don't think a table of contents should be inside the header, probably better to keep only the h2 inside header and then wrap the header and toc in a div if we need a container.

tan-90 (approved):
Left one normal comment and one minor one, please take a look. Everything looks good though so I'm approving in advance.

PR #733, opened 2022-09-19
fix: added botd type to response

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/726


src/components/botd/APIResponseDetailsSection/APIResponseDetailsSection.tsx
```diff
@@ -54,6 +59,25 @@ export default function APIResponseDetailsSection() {
...
+                key='type'
+                className={styles.type}
+                left={
+                  visitorData?.products.botd.data.bot.type
+                    ? 309 + visitorData?.products.botd.data.bot.type.length * characterLength
+                    : 284
+                }
+              >
+                <p>
+                  Name of <strong>automation tool</strong> which is used for the request.
```
tan-90, line 72: Has it been proofread? Feels like it's missing a "Name of _the_ automation tool" and I'd use "which ~is~ was used".

tan-90 (approved):
Approving in advance with a couple minor comments, please take a look.

PR #734, opened 2022-09-20
fix: added reset chatbot widget on selected pages

natapg:
closes https://github.com/fingerprintjs/fingerprint.com/issues/725


src/constants/content.ts
```diff
@@ -143,3 +143,18 @@ export const industryDropdown = [
   { title: 'Online Gaming & Gambling', url: PATH.gaming },
   { title: 'Cryptocurrency', url: PATH.cryptocurrency },
 ]
+
+// We need to reset the chatbot widget on the following pages
+export const PAGES_RELOAD_CHATBOT = [
+  '/use-cases/personalization/',
+  '/use-cases/account-sharing-prevention/',
+  '/use-cases/payment-fraud/',
+  '/use-cases/credential-stuffing/',
```
tan-90, line 152: Optional: We could do `PATH.useCases.<...>`

tan-90 (approved):

src/constants/content.ts
natapg: you mean  `${PATH.useCases}personalization/` right?

PR #750, opened 2022-10-05
contact sales: form section

natapg:
Mockup https://www.figma.com/file/niHrIRowTbqQhDO7W5CPNf/Web-Mockups?node-id=6374%3A46122


src/components/ContactSalesForm/index.tsx
```diff
@@ -96,9 +97,7 @@ export default function ContactSalesForm() {
           updateErrorMessage(error?.message)
         }
       }
-      setTimeout(() => {
-        updateFormState(FormState.Default)
-      }, 10000)
```
tan-90, line 101: Do we not need to reset the form anymore? Wouldn't this mean the error message would get stuck and need a refresh?

tan-90 (approved):
Just one question, please take a look. All looks good, approving in advance.

PR #760, opened 2022-10-11
feat: added advertising pages

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/719


gatsby-node.js
```diff
@@ -223,6 +223,23 @@ exports.createPages = async ({ actions, graphql }) => {
...
+    const newPage = {
+      ...page,
+      context: {
+        ...page.context,
+        noIndex: true,
+      },
+    }
+    deletePage(page)
+    createPage(newPage)
+  }
```
tan-90, line 240: This means all ads pages should be `noIndex` right? Why do we need to recreate the page instead of setting `noIndex` from the start?

src/components/demo/GetStartedSection/GetStartedSection.tsx
```diff
@@ -15,15 +15,25 @@ import { ReactComponent as YahooSVG } from '../../../../static/img/company-logos
...
+        {!advertisingVariant && <h2 className={styles.subTitle}>Get Started For Free Today</h2>}
+        {advertisingVariant ? (
+          <Button size='big' href={PATH.contactSales} className={styles.advertisingButton}>
+            Contact Sales
+          </Button>
+        ) : (
+          <Button size='big' href={URL.signupUrl} className={styles.button}>
+            Create Free Account
+          </Button>
+        )}
```
tan-90, line 36:
```suggestion
        {advertisingVariant ? (
          <Button size='big' href={PATH.contactSales} className={styles.advertisingButton}>
            Contact Sales
          </Button>
        ) : (
          <h2 className={styles.subTitle}>Get Started For Free Today</h2>
          <Button size='big' href={URL.signupUrl} className={styles.button}>
            Create Free Account
          </Button>
        )}
```

src/pages/ads/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { SEO } from '../../components/SEO/SEO'
+import { withPrefix } from 'gatsby'
+import VideoSection from '../../components/VideoSection/VideoSection'
+import ProToolsSection from '../../components/ProToolsSection'
+import StayProtectedSection from '../../components/StayProtectedSection'
+
+export default function IndexPage() {
+  return (
+    <Layout>
+      <div>
```
tan-90, line 18: Can't we remove the extra div and just have it all under the layout?

src/pages/advertising/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { SEO } from '../../components/SEO/SEO'
+import { withPrefix } from 'gatsby'
+import VideoSection from '../../components/VideoSection/VideoSection'
+import ProToolsSection from '../../components/ProToolsSection'
+import StayProtectedSection from '../../components/StayProtectedSection'
+
+export default function IndexPage() {
+  return (
+    <Layout>
+      <div>
```
tan-90, line 20: Same here.

tan-90 (dismissed):
Left some comments, please take a look. The most important one is about not recreating the pages, if it's possible we should change that. Nothing major though so I'm approving in advance.

gatsby-node.js
natapg: those pages are created automatically, so we cannot add context params from the beginning, and according to the documentation this is how to add context 
https://www.gatsbyjs.com/docs/creating-and-modifying-pages/#pass-context-to-pages

tan-90 (approved):

PR #768, opened 2022-10-14
fix: added botd fields

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/767


dcorso21 (approved):
The new fields look good, I think based on the length of the user agent we should consider having the tooltip be on the left. We could also consider having a smaller indent as well. 

Finally, the type parameter shows up as empty for me. maybe it should show "undefined" when no bot is detected. 

image

PR #772, opened 2022-10-18
github: added chart section

natapg:


src/components/githubLandingPage/AccuracySection/AccuracySection.tsx
```diff
@@ -0,0 +1,296 @@
...
+}
+export default function AccuracySection() {
+  const chartWidth = 860
+  const chartHeight = 414
+  const xPoints = 4
+  const maxY = 100000
+  const offset = 25
+  const yPointsPro = [99500, 99003, 98507, 98015]
+  const xPointsLine = ['30 days', '60 days', '90 days', '120 days']
+  const yPointsLine = ['100k', '75k', '50k', '25k', '0']
```
tan-90, line 30: Minor: d3 calls this a scale:

```suggestion
  const xScale = ['30 days', '60 days', '90 days', '120 days']
  const yScale = ['100k', '75k', '50k', '25k', '0']
```

src/components/githubLandingPage/AccuracySection/AccuracySection.tsx
```diff
@@ -0,0 +1,296 @@
...
+                  >
+                    <div
+                      className={styles.lineSegmentOss}
+                      style={
+                        {
+                          '--hypotenuse': HypoOss[index],
+                          '--angle': Math.asin(AngleOss[index]) * (180 / Math.PI),
+                        } as React.CSSProperties
+                      }
+                    />
```
tan-90, line 272: I was thinking about maybe making a component to remove this repetition but the pro list also has the extra divs for the hover events so I think it wouldn't be any cleaner if we extracted some of it to another component.

tan-90 (approved):
Looking great as always, only very minor comments! 🎉

PR #786, opened 2022-10-25
github: layers section

natapg:


dcorso21 (approved):
Looks great! The only thing I found was when in mobile. You can see that the dashed line seems to double for a second when selecting a layer. 

Its not major by any means though, so I'm approving in advance.

githubLayers

PR #811, opened 2022-11-11
top nav bar changes

natapg:


dcorso21 (approved):
Approving, I would consider renaming the PR to relate to the slight style changes applied though.

CC @natapg

PR #815, opened 2022-11-14
fix: bring back message field contact sales form

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/814


dcorso21 (approved):
Approving, although I noticed that there is a change in casing here. Is it intentional. In any case this is out of scope, so it should not block this change.

dcorso21 (approved):

PR #825, opened 2022-11-22
fix: added mumbai region to faq page

natapg:
request : could you add Asia-Pacific(Mumbai) to the below question on the FAQ page?

image


src/pages/resources/frequently-asked-questions-faqs/index.tsx
```diff
@@ -473,8 +473,8 @@ export const faqs = [
         answer: (
           <>
             <p className={styles.faqContent}>
-              When you create your account, you can choose between Global/US data storage (Richmond, Virginia) and data
-              EU storage (Frankfurt, Germany).
+              When you create your account, you can choose between Global/US data storage (Richmond, Virginia), data EU
+              storage (Frankfurt, Germany) and Asia-Pacific storage (Mumbai).
```
dcorso21, line 477: I think that the line should probably be:
 "...choose between Global/US data storage (Richmond, Virginia), EU data storage (Frankfurt, Germany) and Asia-Pacific data storage (Mumbai)"

OR 

 "...choose between Global/US storage (Richmond, Virginia), EU storage (Frankfurt, Germany) and Asia-Pacific storage (Mumbai)"

(with or without data - it should be parallel.)

dcorso21 (changes requested):

dcorso21 (approved):

PR #831, opened 2022-11-27
homepage: for developers section

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/795
Mockup [figma.com/file/fTnEEueuxlkKrubDegp9aw/HP?node-id=775%3A41675&t=4d1tRbc6njRuAd62-0](https://www.figma.com/file/fTnEEueuxlkKrubDegp9aw/HP?node-id=775%3A41675&t=4d1tRbc6njRuAd62-0)
DP: [deploy-preview-831--fingerprintcom.netlify.app](https://deploy-preview-831--fingerprintcom.netlify.app/)


src/components/homepage/ForDevelopersSection/ForDevelopersSection.tsx
```diff
@@ -0,0 +1,283 @@
...
+        <header className={styles.header}>
+          <div className={styles.labelWrapper}>
+            <span className={styles.label}>For developers</span>
+          </div>
+          <h2 className={styles.title}>Fingerprint is easy to implement</h2>
+          <p className={styles.description}>
+            Developer teams can get started quickly with our intuitive APIs and extensive documentation.
+          </p>
+        </header>
+        <div className={styles.cards}>
```
tan-90, line 59: For creating the cards wouldn't it be cleaner to move the non-common props into the `cards` object and then just map over `Object.values(cards)`? That way we would only need to write `<Card ...` or `<MobileCard ...` once.

src/components/homepage/ForDevelopersSection/ForDevelopersSection.module.scss
```diff
@@ -0,0 +1,333 @@
...
+      background: linear-gradient(94.48deg, #f5614b 0%, #f5614b 0.01%, #fa7545 100%);
+      -webkit-background-clip: text;
+      -webkit-text-fill-color: transparent;
+      background-clip: text;
+      text-fill-color: transparent;
+
+      @include media('<=desktopHeader') {
+        font-size: rem(12px);
+      }
+    }
```
tan-90, line 56: These gradient text colors are getting used quite a bit, maybe we could extract it into a mixin?

tan-90 (approved):
Left a minor style suggestion about rendering the cards, we can discuss that one.
I think the gradient text mixin is something we should do but that one could be done in a separate PR.
All else looks good!

PR #839, opened 2022-11-30
fix: added calendar link based on region

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/838


src/components/ContactSalesForm/index.tsx
```diff
@@ -93,8 +96,26 @@ export default function ContactSalesForm() {
...
+            break
+          case Region.APAC:
+            calendarUrl = URL_CALENDAR.contactSalesCalendarApac
+            break
+          case Region.EMEA:
+            calendarUrl = URL_CALENDAR.contactSalesCalendarEmea
+            break
+          default:
+            calendarUrl = URL_CALENDAR.contactSalesCalendarAmer
+        }
```
dcorso21, line 115: nit: I would probably use a map here with keys for each region and values for the calendars, with the fallback being the american one.

src/components/ContactSalesForm/index.tsx
```diff
@@ -93,8 +96,26 @@ export default function ContactSalesForm() {
       } else {
         updateFormState(FormState.Success)
         trackLeadSubmit()
+        let calendarUrl: string
+        switch (countryRegion) {
+          case Region.AMER:
+            calendarUrl = URL_CALENDAR.contactSalesCalendarAmer
+            break
```
dcorso21, line 103: If you put this at the end, you can use
```ts
case Region.AMER:
default:
    calendarUrl = URL_CALENDAR.contactSalesCalendarAmer
```

src/components/ContactSalesForm/index.tsx
dcorso21: as in 
```ts
const calendarUrl = calendarMap[countryRegion] || URL_CALENDAR.contactSalesCalendarAmer
```

dcorso21 (approved):
Looks great! I didn't find any issues, but I left a couple code formatting suggestions below - totally optional.

PR #1030, opened 2023-01-20
feat: prepare to gatsby v5 update

natapg:
1. Update Gatsby to the latest v4 version
2. Update CI to use Node v18
3. Added id to Script
4. fixed validateDOMNesting issues like `<a> cannot appear as a descendant of <a>`
5. The $alpha argument must be unitless and between -1 and 1 (inclusive).
6. Added slidesPerView='auto' to Swiper.js to fix this issue https://github.com/nolimits4web/swiper/issues/5784#issuecomment-1157783235
7. Added key to column iterable when there is no title
8. removed react-onclickoutside package and added a custom hook
9. removed FC from test-utils
10. optimize images


tan-90 (approved):
Please take a look at the requested changes, but they're very minor so I'm approving in advance.

PR #1050, opened 2023-01-31
Chore: update to Gatsby 5 / React 18

natapg:
Update to Gatsby v5 including all Gatsby dependencies.
Update to React 18.
Update to framer-motion v9.
Update to simple-oauth2 v5.
Migrate Gatsby files from .js to .mjs to [use ESM imports](https://github.com/gatsbyjs/gatsby/discussions/37069).
Migrate from Remark to [Micromark](https://github.com/micromark/micromark) which is [recommended by Remark](https://github.com/remarkjs/remark#when-should-i-use-this) in our use case.
Moved console promotional message to gatsby browser so it runs only once


gatsby-node.mjs
```diff
@@ -406,15 +408,18 @@ exports.sourceNodes = async ({ actions, getNodes, createContentDigest }) => {
...
     name: 'md',
     extend() {
       return {
         resolve(source, args, context, info) {
           const fieldValue = context.defaultFieldResolver(source, args, context, info)
-          return remark().use(remarkHTML).processSync(fieldValue).toString()
+          if (fieldValue) {
+            return micromark(fieldValue)
+          }
+          return
```
tan-90, line 422:
```suggestion
          return ''
```

tan-90, line 422:
```suggestion
          return ''
```

tan-90 (approved):

PR #1081, opened 2023-02-14
feat: add region-specific logos

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/1054

[figma.com/file/Ad7wlJvqIMSOwWT5ung0Or/Website-New?node-id=1154%3A16372&t=GZ0MrJ83VFy9p6ZJ-1](https://www.figma.com/file/Ad7wlJvqIMSOwWT5ung0Or/Website-New?node-id=1154%3A16372&t=GZ0MrJ83VFy9p6ZJ-1)


src/components/homepage/ClientsSection/ClientsSection.tsx
```diff
@@ -17,48 +17,131 @@ import { ReactComponent as TripleWhaleSVG } from '../../../../static/img/company
...
+      <RockstarSVG className={classNames(styles.logo, styles.rockstar)} />
+      <YahooSVG className={classNames(styles.logo, styles.yahoo)} />
+      <TargetSVG className={classNames(styles.logo, styles.target)} />
+      <Upstar className={classNames(styles.logo, styles.upstar)} />
+      <CheckoutSVG className={classNames(styles.logo, styles.checkout)} />
+      <TripleWhaleSVG className={classNames(styles.logo, styles.tripleWhale)} />
+    </>
+  )
+
+  const [regionLogos, setRegionLogos] = useState(amerLogos)
```
tan-90, line 70: Maybe we should start it empty and just reserve the space for the logos, then only render after we know the region. Right now the american logos will flash before it switches to the correct ones once the visitor data is loaded.

tan-90 (changes requested):
The code looks good but the logos flash which doesn't look great, I think we should change it to only render once we're sure we have the correct region.

src/components/homepage/ClientsSection/ClientsSection.tsx
```diff
@@ -104,43 +103,50 @@ export default function ClientsSection() {
...
-      case Region.APAC:
-        setRegionLogos(apacLogos)
-        break
-      case Region.EMEA:
-        setRegionLogos(emeaLogos)
-        break
-      default:
-        break
+    if (userCountry) {
+      switch (countryRegion) {
```
tan-90, line 107: Minor: can't we check `if (countryRegion)` directly?

tan-90 (approved):

src/components/homepage/ClientsSection/ClientsSection.tsx
natapg: no, because it already has a default value

PR #1118, opened 2023-03-01
fix: added clear timeouts

natapg:


src/components/homepage/HeroSection/HeroSection.tsx
```diff
@@ -32,6 +32,7 @@ export default function HeroSection({ advertisingVariant = false }: HeroSectionP
       ref.current.play()
     }
   }, [isInView])
+  let timer: NodeJS.Timeout
```
tan-90, line 35: This declaration should be inside the effect. It works the same since we pass it to a closure but react will complain because the assignment gets lost when the component re-renders.

tan-90 (approved):
Approving in advance with a small change request, please take a look.

PR #1124, opened 2023-03-02
chore: update Swiper to V9

natapg:
Updating to v9 to use native browser lazy loading instead of Swiper lazy component


src/components/homepage/DemoSection/DemoSection.module.scss
```diff
@@ -415,12 +415,22 @@
...
             }
           }
-          .mapImage,
-          .mapSkeleton {
+          .mapImage {
             object-fit: none;
             width: rem(112px);
             height: rem(96px);
             border-radius: rem(4px);
+            background-color: #f2f2f7;
```
dcorso21, line 423: nit for later: looks like we use this color often. Maybe we should make it a variable later?

dcorso21 (approved):

PR #1162, opened 2023-03-15
fix: minified errors

natapg:
Changes to fix minified errors like 
hydration failed because the initial ui does not match what was rendered on the server:

For this I have made the following changes:

- round values that are used in the homepage table. For example we currently have angles like angle:13.1760052754384, and that sometimes varies between server and client rendering for some reason.
- added suppressHydrationWarning to elements that use .toLocaleString() or new Date(timestamp) that can also make between server and client rendering
- removed browser checks from useEffect because is not necessary to have them there


dcorso21 (approved):
Thanks for the explanation. I think it makes sense to me.

PR #1166, opened 2023-03-15
fix: apac logos

natapg:
From Valentin
This logo has a line artifact at the top, is it intentional or we should remove it?

image

Some newer logos look smaller, compared to older ones, e.g. if you compare "uni" or "happyfresh" with "Agoda", they look about 80% smaller, wanted to clarify if this is intentional.


dcorso21 (approved):
LGTM!

PR #1167, opened 2023-03-15
fix: remove gatsby-plugin-react-svg

natapg:
removing gatsby-plugin-react-svg since it uses a vulnerable version of loader-utils
(also Babel to clean this warning https://github.com/babel/babel/pull/15484)


gatsby-node.mjs
```diff
@@ -312,12 +312,32 @@ export const onCreateWebpackConfig = ({ stage, actions, getConfig }) => {
       rules: [
         {
           test: /\.(mov)$/,
-          loader: 'file-loader',
-          options: {
-            name: '[name].[hash].[ext]',
-            publicPath: './assets',
-            outputPath: './assets',
-          },
+          type: 'asset/resource',
```
natapg, line 315: since we use Webpack 5, changing to asset modules
https://webpack.js.org/guides/asset-modules/

dcorso21 (approved):

PR #1174, opened 2023-03-17
fix: wrap visitors endpoint

natapg:
wrap the visits endpoint as we did with the events endpoint to prevent it from being blocked again in the future


api/redirect/visits.ts
```diff
@@ -0,0 +1,21 @@
...
+export default async function visitsEndpoint(req: VercelRequest, res: VercelResponse) {
+  // Get the visitorId parameter from the request body
+  const { visitorId } = req.body
+
+  const endpoint = process.env.GATSBY_FPJS_VISITORS_ENDPOINT
+  const key = process.env.GATSBY_FPJS_SECRET_TOKEN
+  try {
+    // Make a request to the events API with the requestId parameter
+
+    const response = await fetch(`${endpoint}${visitorId}?token=${key}&limit=21`)
```
dcorso21, line 13: This may be a good opportunity to update the query param from `token` to `api_key` which is the term used in our docs since the name change.

dcorso21: I would also suggest preserving the comment which states that the 21 limit is for getting 20 events.

dcorso21 (approved):
It's so cool to see how simple it is to proxy this request! It makes me think of all the ways we may use it in the future. 

I've left some minor comments, but it is up to you -  looks good. Thanks!

PR #1177, opened 2023-03-20
fix: move dateFormatter and numberFormatter to factory function

natapg:
moved functions to factory function to avoid hydration problems


dcorso21 (approved):
LGTM!

PR #1204, opened 2023-03-27
fix: bot detection changes

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/1182


src/components/botd/FeaturesSection/BrowserSVG.svg
tan-90, line 1: image

src/pages/products/bot-detection/index.tsx
```diff
@@ -80,15 +80,15 @@ export default function Botd({ pageContext }: BotdProps) {
         githubLink={URL.botDRepoUrl}
         labels={['+20K downloads', '+58M API requests']}
       >
-        BotD is an open source project supported by contributing developers across the globe.
+        Bot detection is an available open source project supported by contributing developers across the globe.
```
tan-90, line 83: Did we proofread this? I'm not sure about the grammar but it sounds a bit weird to me.

tan-90 (approved):

src/pages/products/bot-detection/index.tsx
natapg: it was written by @jackspirou and @Kroy915 so I think so

PR #1220, opened 2023-04-04
fix: check if homepage video can be auto-played (chrome)

natapg:
based on 
https://developer.chrome.com/blog/autoplay/


tan-90:
Can this happen under normal circumstances? The article mentions `.pause` as the cause but I don't think we call pause anywhere.

src/components/homepage/HeroSection/HeroSection.tsx
```diff
@@ -27,11 +29,23 @@ export default function HeroSection({ advertisingVariant = false }: HeroSectionP
...
+
   useEffect(() => {
     if (isInView && ref.current) {
-      ref.current.play()
+      const playVideo = async () => {
+        if (ref.current) {
+          try {
+            await ref.current.play()
+          } catch (error) {
+            rollbar.warning('Auto-play was prevented', getErrorMessage(error))
```
tan-90, line 41: This depends on the answer to the question I left on the PR, but if it's a client only error it might not be necessary to log it to rollbar.

tan-90 (commented):
Left a question and a comment, I want to discuss this a bit more.

src/components/homepage/HeroSection/HeroSection.tsx
tan-90: > Can this happen under normal circumstances? The article mentions `.pause` as the cause but I don't think we call pause anywhere.

Discussed it on slack, autoplay being prevented also makes `.play` throw.

tan-90 (approved):

src/components/homepage/HeroSection/HeroSection.tsx
natapg: thanks Vini :)

PR #1221, opened 2023-04-04
fix: demo for browsers without SessionStorage support

makma:
The `SessionStorage` is not available in some browsers which breaks demo for some users. Fingerprint Pro React lib [uses](https://github.com/fingerprintjs/fingerprintjs-pro-react#caching-strategy) the `SessionStorage` as a default caching strategy.

Please note, I'm not fully familiar with this website and the recommended fix needs a proper review and testing since I don't know if something relies on this cache (I was not able to find anything).

### DIFF
--- a/src/AppProviders.tsx
+++ b/src/AppProviders.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import { FormProvider } from './hooks/useForm'
 import { FpjsProvider } from '@fingerprintjs/fingerprintjs-pro-react'
 import * as FingerprintJS from '@fingerprintjs/fingerprintjs-pro'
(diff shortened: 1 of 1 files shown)


natapg:
@makma session storage has high browser compatibility, could you give me an example of where it doesn't work? In such a case, it shouldn't break, it should just stop using cache, and call the api call on every page change instead of checking session storage, can that be "fixed" in Pro React?

ilfa:
I'll add a fallback to the SPA package and it will fix the problem with a better way.

natapg:
thanks @ilfa @makma , I'll close this PR and we can discuss it later

PR #1233, opened 2023-04-11
fix: Add a hint what to try when use case Stackblitz demo is broken

JuroUhlar:
Hey! :) 

I would like to propose adding a line underneath our Stackblitz demos with a hint on how to get it to work if the visitor uses Firefox/Safari or strict ad-blockers.

Context: https://fingerprintjs.slack.com/archives/C02P7D7TCS3/p1681136588437769

Thanks in advance!


natapg:
looks good! please add some padding between the iframe and the hint (8px), thanks! Approving in advance

PR #1293, opened 2023-05-06
fix: button component refactor

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/810


tan-90 (approved):
Nice refactor, the prop names are a lot more clear.

PR #1316, opened 2023-05-18
feat: add fuzzer

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/1305


src/pages/products/bot-detection/index.tsx
```diff
@@ -35,18 +43,65 @@ export default function Botd({ pageContext }: BotdProps) {
   const { visitorData, requestId } = useBotDContext()
   const rollbar = useRollbar()
 
+  const runFuzzer = async () => {
+    const IGNORE_KEYS = [
+      // Useful for reducing payload size
+      new RegExp(
+        'window.(modernizr|webpack|prism|dataLayer|google_tag_data|__adroll|___grecaptcha_cfg|closure_lm|rollbar|_rollbar|asyncRequires|__vercel_toolbar|gaData|grecaptcha|google_tag_manager|recaptcha|GoogleAnalyticsObject|regeneratorRuntime|___loader|gaGlobal|gaplugins|adroll*|analyticsConnectorInstances|__core-js*|___emitter*|__google_recaptcha_*|__fpjs_*|__recaptcha*)+',
```
ctizen, line 50: @xnerhu 
Does this regexp really do what it was meant to do? What I can see here:
- Unescaped dot after `window` would match any symbol
- Weird `*` quantifiers inside alternation group (e.g., `__core-js*` would match `__core-jssssssssss...` 🐍)
- Ending `+` quantifier with unclear purpose.

Could you please clarify what should it do? :)

xnerhu: I fixed unscaped dot.

xnerhu: `+` ending requires at least one pattern match

With `*` I wanted to match exactly start of string

ctizen: @xnerhu 

> `+` ending requires at least one pattern match
Why would you need more than one match here? It looks like it should be exactly one, so the quantifier should be removed.

> With `*` I wanted to match exactly start of string
Asterisk is a zero-or-more quantifier, it doesn't work this way. Regexps and globs are different things.

I'd suggest adding some jest tests for this regexp and/or test it with some online tool like https://regex101.com - just to make sure it really behaves the way you want it to.

ctizen: OK, I can approve this in current version :)

PR #1328, opened 2023-05-24
Fix/use vercel functions

natapg:


dcorso21 (approved):
Looks good, nice that we can keep it out of the request.

PR #1350, opened 2023-05-31
Smart Signals: Identification page

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/1333

basically it is the [pro](https://fingerprint.com/products/fingerprint-pro/) page but with some changes

DP:https://fingerprint-com-git-feat-integrations-page-fp-pro.vercel.app/products/identification/

Figma:https://www.figma.com/file/Ad7wlJvqIMSOwWT5ung0Or/Website-New?type=design&node-id=2941%3A70725&t=1HFqdSbYCEU4Vvai-1


src/components/AccuracySection/DiagramSection/DiagramSection.tsx
```diff
@@ -96,9 +110,10 @@ interface TabsProps {
   currentTab: tabOptions
   setCurrentTab: (tabOption: tabOptions) => void
   className: string
+  fingerprinting?: boolean
```
tan-90, line 113: I'd change this to `product?: 'Fingerprinting' | 'Pro'` and have "Pro" as a default value, that way it will be reusable for other products by just changing the type.

src/components/identification/HeroSection/HeroSection.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { motion, AnimatePresence } from 'framer-motion'
+
+import styles from './HeroSection.module.scss'
+
+export default function HeroSection() {
+  const [render, isRender] = useState(false)
+
+  useEffect(() => {
+    isRender(true)
+  }, [])
```
tan-90, line 16: Minor: this could be a hook

src/constants/env.ts
```diff
@@ -56,7 +56,9 @@ export const ROLLBAR_ACCESS_TOKEN = process.env.GATSBY_ROLLBAR_ACCESS_TOKEN ?? '
 export const BRANCH = process.env.BRANCH
 
 export const HOST =
-  process.env.GATSBY_VERCEL_ENV === 'preview' ? `https://${process.env.GATSBY_VERCEL_URL}` : 'https://fingerprint.com'
+  process.env.GATSBY_VERCEL_ENV === 'preview' && typeof window !== 'undefined'
+    ? `https://${window.location.hostname}`
+    : 'https://fingerprint.com'
```
tan-90, line 61: I don't remember if we do SSR or SSG, this might cause problems with SSG since at the time of generation the window will always be undefined.

tan-90 (approved):

tan-90:
Just a random related comment, we could use `height: 100vh` to make the first section takes up all vertical space and that way we'd not see the background from the next section. It can make resizes look weird and it might not be clear that there's more content to see so it's not always a good choice.

Screen Shot 2023-05-31 at 13 53 16

PR #1355, opened 2023-05-31
Smart Signals: platform page

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/1349

basically it is the [pro](https://fingerprint.com/products/fingerprint-pro/) page but with some changes

DP:https://fingerprint-com-git-smart-signals-platform-page-fp-pro.vercel.app/products/fingerprint-pro/

Figma:https://www.figma.com/file/Ad7wlJvqIMSOwWT5ung0Or/Website-New?type=design&node-id=3128%3A63220&t=q1bVOfpKu9UBcCki-1


src/components/pro/CardSection/CardSection.tsx
```diff
@@ -9,13 +9,11 @@ export default function CardSection() {
...
-          </section>
+          <h2 className={styles.title}>The internet relies on Fingerprint</h2>
           <section className={styles.labelsSection}>
-            <Card value='6k+' field='companies' />
-            <Card value='16%' field='of top 500 sites' />
-            <Card value='1.2B+' field='API calls/month' />
+            <Card value='6K+' field='companies' />
+            <Card value='1.2B+' field='API calls/month' valueMobile='16%' fieldMobile='of top 500 sites' />
+            <Card value='16%' field='of top 500 sites' valueMobile='1.2B+' fieldMobile='API calls/month' />
           </section>
```
tan-90, line 17: You should use [the order property](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items#the_order_property) instead.

tan-90 (approved):
There's a very small change request but everything else looks good :bongo-blob:

PR #1397, opened 2023-06-09
fix: remove target logo from website

natapg:


src/pages/ads/demo/index.tsx
```diff
@@ -35,7 +34,6 @@ export default function DemoPage({ pageContext }: DemoPageProps) {
       <DemoSection />
       <CatchFraudstersSection />
       <SolveAnyFraudSection />
-      <GetStartedSection advertisingVariant />
```
natapg, line 38: the advertising variation was outdated from the original demo so removing this section

bwegrzyn (approved):

PR #1417, opened 2023-06-14
fix: add empty state pricing

natapg:
Closes https://github.com/fingerprintjs/fingerprint.com/issues/1396 and https://github.com/fingerprintjs/fingerprint.com/issues/1408

I changed the table element for a div to be able to have more control of the animations and to be able to add the accordion effect


dcorso21 (approved):
LGTM!

PR #1438, opened 2023-06-26
chore: update gatsby deps

natapg:


package.json
```diff
@@ -39,42 +39,42 @@
...
-    "html-entities": "^2.3.5",
+    "gatsby-source-filesystem": "^5.11.0",
+    "gatsby-transformer-remark": "^6.11.0",
+    "gatsby-transformer-sharp": "^5.11.0",
+    "gatsby-transformer-yaml": "^5.11.0",
+    "html-entities": "^2.4.0",
     "include-media": "^2.0.0",
     "jest-environment-jsdom": "^29.0.3",
-    "micromark": "^3.2.0",
+    "micromark": "^4.0.0",
```
tan-90, line 77: Since everything is green I'm guessing none of the breaking changes on this major version hit APIs that we use. If you didn't already it's worth doing a quick check if markdown is still working as expected.

tan-90 (approved):
Small reminder to double check a major version bump but approving in advance since the build is green.

package.json
natapg: Yes, I checked everything and it works as expected, there is no big change, just that we have to use node 18, which we have already been using

PR #1460, opened 2023-07-07
fix: wait for botd response to send ampli event

natapg:
image


dcorso21 (approved):
Ah, I see, sorry I didn't catch it before. Looks good to me!

PR #1490, opened 2023-07-19
fix: refactor color variable names for clarity

natapg:
`github-white` ->`warmWhite` 

`github-orange`, `github-orange-gradient`, `github-orange-dark`: The prefix "github-" was removed from these variable names

`lightGrey` with value `#d0d0d0`, was added


tan-90 (approved):
These are always good to see, great work on the refactor!

PR #1546, opened 2023-08-15
feat: add amplitude experiments

natapg:


src/pages/index.tsx
```diff
@@ -17,10 +17,69 @@ import BadgesSection from '../components/homepage/BadgesSection/BadgesSection'
...
+
+    sendExposureEvents()
+  }, [visitorId, serverData, rollbar])
+
+  const heroVariant = getValidExperimentValue(serverData.variants[ExperimentFlagHomepage.HeaderExperimentTest]?.value)
+
   return (
     <LayoutTemplate>
-      <HeroSection />
+      <HeroSection variant={heroVariant} />
```
spalt08, line 82: nit: probably OK for the first experiment, however in the future I would suggest using react context for experiment flags, to avoid prop drilling

src/pages/index.tsx
```diff
@@ -40,3 +99,75 @@ export default function IndexPage() {
...
+
+export async function getServerData(req: GetServerDataProps) {
+  const headers: { [key: string]: string[] } = {
+    'Set-Cookie': [],
+  }
+  const cookieHeader = req.headers.get('cookie') || ''
+  const parsedCookies = cookie.parse(typeof cookieHeader === 'string' ? cookieHeader : '')
+
+  let userId = parsedCookies.ampli_experiment_server_user_id
+  const storedFlags = parsedCookies.ampli_stored_flags ? JSON.parse(parsedCookies.ampli_stored_flags) : {}
```
spalt08, line 111: I don't think we should cache flags in cookies to be honest. I would suggest to re-fetch flags on page refresh. It should be consistent because we rely on `server_user_id` which is cached. If we face performance issues because of this, we can re-think the implementation and add some caching later

src/pages/index.tsx
```diff
@@ -40,3 +99,75 @@ export default function IndexPage() {
 export function Head() {
   return <SEO />
 }
+
+export async function getServerData(req: GetServerDataProps) {
+  const headers: { [key: string]: string[] } = {
+    'Set-Cookie': [],
+  }
+  const cookieHeader = req.headers.get('cookie') || ''
```
spalt08, line 107: nit: I would suggest to extract all cookie-related logic into a separate function to improve code readability

src/pages/index.tsx
```diff
@@ -40,3 +99,75 @@ export default function IndexPage() {
...
+      })
+    )
+  }
+
+  // Initialize the experiment.
+  const experiment = Experiment.initializeRemote(AMPLITUDE_EXPERIMENT_DEPLOY_API_KEY)
+
+  const user = {
+    user_properties: {
+      websiteABTestKey: userId,
```
spalt08, line 139: nit: we can rename the property to something like `serverSideCookieId` and re-use this in the future experiments

api/redirect/amplitude-experiments.ts
```diff
@@ -0,0 +1,47 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node'
+import { getErrorMessage } from '../_utils/utils'
+
+export default async function amplitudeExperiments(req: VercelRequest, res: VercelResponse) {
```
spalt08, line 4: nit: we should probably rename the function and the path to something like `amplitude experiment exposure`

src/pages/index.tsx
natapg: I've removed the flag logic

src/pages/index.tsx
natapg: extracted to a separate function

api/redirect/amplitude-experiments.ts
natapg: renamed

src/pages/index.tsx
natapg: renamed to serverSideCookieId

src/pages/index.tsx
```diff
@@ -17,10 +17,75 @@ import BadgesSection from '../components/homepage/BadgesSection/BadgesSection'
...
+  const rollbar = useRollbar()
+
+  useEffect(() => {
+    if (serverData.error) {
+      rollbar.error('Error fetching experiment variants', serverData.error)
+      return
+    }
+  }, [serverData, rollbar])
+
+  useEffect(() => {
```
spalt08, line 57: nit:
```js
useEffect(async () {
  if (!visitorId || !serverData) {
     return
  }
  // ...
}, [visitorId, serverData, rollbar])

spalt08 (approved):

PR #1555, opened 2023-08-17
fix: use local evaluation for amplitude experiments

natapg:


src/helpers/amplitude-experiment.ts
```diff
@@ -35,17 +35,15 @@ export async function fetchExperimentVariants(
     )
   }
 
-  // Initialize the experiment.
-  const experiment = Experiment.initializeRemote(AMPLITUDE_EXPERIMENT_DEPLOY_API_KEY)
+  // Initialize the experiment for local evaluation.
```
spalt08, line 38: I would rather do something like:

```
const experiment = Experiment.initializeLocal(AMPLITUDE_EXPERIMENT_DEPLOY_API_KEY)
const experimentPromise = experiment.start()

export async function fetchExperimentVariants(
  req: GetServerDataProps,
  flags: string[]
): Promise<FetchExperimentVariantsResponse> {
  // ...
  await experimentPromise
  // ...
}
```

spalt08 (approved):

PR #1573, opened 2023-08-22
feat: a/b experiments using edge functions

natapg:


middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+    url.pathname = bucket === 'control' ? '/home-a/' : '/home-b/'
+  } else if (/\/page-data\/home-(a|b)\/page-data\.json/.test(url.pathname)) {
+    url.pathname = bucket === 'control' ? '/page-data/home-a/page-data.json' : '/page-data/home-b/page-data.json'
+  } else {
+    return next()
+  }
+  const headers: HeadersInit = {}
+
+  if (setCookie) {
```
spalt08, line 30: nit: `setCookie` is redundant since it's only used in 1 place

src/templates/home-a.tsx
```diff
@@ -14,10 +14,42 @@ import SOCSection from '../components/homepage/SOCSection/SOCSection'
...
+import { useVisitorData } from '@fingerprintjs/fingerprintjs-pro-react'
+import { getConfig } from '../helpers/fpjs'
+import { sendExposureEvent } from '../helpers/api'
+import { useRollbar } from '@rollbar/react'
+import { getErrorMessage } from '../helpers/error'
+import { ExperimentFlags, ExperimentValue } from '../helpers/amplitude'
 
 import { SEO } from '../components/SEO/SEO'
 
 export default function IndexPage() {
```
spalt08, line 26: Can we create `home.tsx` with regular `IndexPage` component?

home-a.tsx:
```js
export default function IndexPageA() {
   useEffect(/** send exposure **/) 
   
   return <IndexPage variant={ExperimentValue.Control} 
}
```

middleware.ts
```diff
@@ -0,0 +1,34 @@
+import { rewrite, next } from '@vercel/edge'
+import { parse } from 'cookie'
+
+export const config = {
+  matcher: ['/', '/page-data/home-a/page-data.json', '/page-data/home-b/page-data.json'],
+}
+
+export default async function middleware(req: Request) {
```
spalt08, line 8: Could you please leave a comment why do we need this middleware and where to configure it?

middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+  const url = new URL(req.url)
+  const cookies = parse(req.headers.get('cookie') ?? '')
+  const bucketName = 'ampli_experiment_server_user_id'
```
spalt08, line 11: `bucketName` should be something like `homepage-header-a-b`

src/templates/home-a.tsx
natapg: I'll update the name

middleware.ts
natapg: yes!

middleware.ts
natapg: you are right, we are not using user_id with this approach

